PR #1, opened 2023-03-02
Update tracked files and gitignore

AmLibra:


vrooomer (approved):
There is nothing to review, we just choose to delete the .idea folder so each contributor uses their own .idea.
With the same intention, we added a line to the .gitignore file so that this folder is never added again.
-v:)

PR #7, opened 2023-03-03
Bootcamp/firebase restore

Jharaxus:
Add firebase support to the project


Kaz-ookid (commented):
PR looks good with clean and clear code. However, as the most part of it will not be of any use for our app (FirebaseActivity and tests), the team decided to close this PR. Furthermore it is still using the basic xml instead of Jetpack Compose, so it seems better to re-write the very few needed line of codes in a new branch.

Nice work!

PR #8, opened 2023-03-03
Add OpenStreetMap map

jadtl:
This adds an OpenStreetMap map using osmdroid, the map activity displays the EPFL campus at start. Internet permission had to be added to `AndroidManifest.xml` for the map to work.


jadtl:
I updated the activity to use Compose :)

app/src/main/java/com/github/polypoly/app/MapActivity.kt
```diff
@@ -0,0 +1,68 @@
+package com.github.polypoly.app
+
+import android.os.Bundle
+import androidx.activity.ComponentActivity
+import androidx.activity.compose.setContent
+import androidx.appcompat.app.AppCompatActivity
```
Jharaxus, line 6: According to AndroidStudio this last import is useless and can be removed

Jharaxus (changes requested):
A little detail to change but apart from that it looks good to me!

Jharaxus (approved):
Nice this fix the previous issue, everything looks fine now.

PR #9, opened 2023-03-03
Bootcamp/kotlin

vrooomer:
This branch corresponds to our migration from Kotlin and the basic XML UI gestion to Kotlin & Jetpack compose.
For this purpose, I needed to recreate the project from scratch.

In this version, we have the classic bootcamp working, i.e. a place where to put you name (can't be empty and max. 15 characters) and a button that shows a screen with a personalized greeting message.

Both two activities were tested (with Jetpack compose & Espresso) in the same fashion of the classic bootcamp.

If we are 100% sure to migrate to Compose, this request should be carefully reviewed and understood as this will the base of our project.

If you have any comments, feel free to ask me!

~ v:)


settings.gradle
```diff
@@ -14,18 +14,3 @@ dependencyResolutionManagement {
 }
 rootProject.name = "polypoly"
 include ':app'
-
-ext.isCiServer = System.getenv().containsKey("CIRRUS_CI")
```
jadtl, line 18: Thanks for the PR :)
These lines should be kept for the CI to run correctly, and the PR is also missing the `.cirrus.yml` and `.codeclimate.yml` files.

jadtl (changes requested):
Thanks for the PR :)
Nitpick: the package name for the project should be something like `com.github.polypoly.app` instead of `com.github.vroomer.polypoly`

vrooomer:
Thanks for the comments, I fixed everything :)

- v:)

jadtl (approved):
LGTM, thanks :)

PR #10, opened 2023-03-07
Google auth

Kaz-ookid:
This PR adds a button to the Main activity that leads to a Login activity where the google authentication system has been implemented. All the authentication takes place in LoginActivity. 

This PR is likely not to be merged, as the LoginActivity is just a usage example of the google authentication system. We shall just come back to this PR and see how it was implemented when the time to implement authentication in our app will be.

We get basic data about the users that logged in, in the Firebase DB (Authentication tab).


vrooomer:
As mentioned by @Kaz-ookid, this PR won't be taken in the project as it doesn't fit in the project and is written in XML (We now work with Jetpack Compose).

However the code seems to be globally fine, thanks!

- v:)

PR #11, opened 2023-03-09
Feat: Add Walk Distance Tracking

AmLibra:
Using Android API to track distance walked. This doesn't add that much code, and even less testable code (to the extent of my abilities). I am open to refactoring and or testing suggestions!

All I can state for sure is that it seems to function well on an emulator!

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -58,6 +58,7 @@ dependencies {
     implementation "androidx.compose.ui:ui:$compose_ui_version"
     implementation "androidx.compose.ui:ui-tooling-preview:$compose_ui_version"
     implementation 'androidx.compose.material:material:1.3.1'
+    implementation 'com.google.android.gms:play-services-location:21.0.1'
     testImplementation 'junit:junit:4.13.2'
     androidTestImplementation 'androidx.test.ext:junit:1.1.5'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.5.1'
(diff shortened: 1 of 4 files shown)


Kaz-ookid (approved):
Code looks clean and passes all checks, well done! 
Functionality works as intended for me. As this PR only adds a basic print in the MainActivity, we can merge the feature with no issues to fear!

vrooomer:
The code itself seems fine as @Kaz-ookid said.

However, as MainActivity.kt is a common activity, I strongly encourage you to not modify it as it could create conflicts.
Also, the logic shouldn't be part of that class neither.

Having a new activity or class would be way better for compatibility and logic purposes (and having tests for the new class too).

I take this review as an opportunity to say to all the members of the project to avoid touching the MainActivity or the GreetingActivity unless it's for local tests.

If you (the team) disagree with the stated points, I would like to hear your propositions!

- v:)

vrooomer (changes requested):
The code itself seems fine as @Kaz-ookid said.

However, as MainActivity.kt is a common activity, I strongly encourage you to not modify it as it could create conflicts.
Also, the logic shouldn't be part of that class neither.

Having a new activity or class would be way better for compatibility and logic purposes (and having tests for the new class too).

I take this review as an opportunity to say to all the members of the project to avoid touching the MainActivity or the GreetingActivity unless it's for local tests.

If you (the team) disagree with the stated points, I would like to hear your propositions!

~ v:)

AmLibra:
While agree that in general:

> Having a new activity or class would be way better for compatibility and logic purposes (and having tests for the new class too).

I deemed adding an extra activity that would later be removed from the project useless, especially if it was going to contain a single block of text.
This decision was further strengthened by the non-test-ability of the added feature, besides messing around with the emulator's location.

I think that a class that would contain all the location related tools could be considered as that library grows. I can always store the @Composable LocationLogic() block in a new class if there are further disagreements!

- Kha

vrooomer (approved):

vrooomer:
Fair enough!

~ v:)

PR #13, opened 2023-03-09
Feature/join group from code

Kaz-ookid:
This PR is not to be merged, as tests have not been implemented yet and it is not fully implemented yet.
PR adds a button in **MainActivity** to launch the **JoinGroupActivity**.
**JoinGroupActivity** enables to input a code to check. If a valid code is entered, it launches a validation toast. We will want to make it launch the group lobby-room activity instead.

It uses a mock DB of valid codes, and functions that checks the code validity are thus not fully implemented. We will want to link with our real Firebase DB and change a few lines in the existing functions to this extent.

All the future changes are marked as TODOs


AmLibra (changes requested):
Good start for the feature implementation. Code is readable, well documented and well encapsulated, as highlighted by the codeclimate score.
Will request changes so that you can finish implementing this feature !

AmLibra (approved):

Kaz-ookid:
Added tests with good coverage, changed Toasts utilisation to composable solution to ease tests and cleaned the code.

Feature still uses mock functions that are marked as TODO's (some tests also are hard coded, waiting for the DB queries to be handled)

I implemented the fact that the "return" button on the emulator keyboard acts as the input button. Ask me if you wish to do it later elsewhere.

PR #14, opened 2023-03-10
Feat: Welcome screen

vrooomer:
### CHECKS:
In this PR I focused myself on one activity: `WelcomeActivity`, as it is purely a UI class (i.e. with no logic at all), I did not write any test (let me know if you have an idea or if I should).

The reason why the coverage is that low is because of the lack of tests mentioned above as the created code duplicates the initial number of code.

Finally, the errors indicated by code climate refer to code that is temporary, so no error is detected on the "official" code.

### GLOBAL UI:
This activity displays a bunch of buttons and images, so far the used images are replace with a mock one `tmp_happysmile` :)
Also, I want to precise that no aesthetic design is fixed now as we're gonna decide on this aspect of the app later (the dark mode isn't supported yet neither).

Moreover, the activity isn't directly accessible from the main one, so to test it you should run the preview `WelcomePreview()` at the end of the activity class.

### RULES:
All the buttons are useless unless the first `OptionButton` (the one at the bottom left) that displays rules, rules that are empty for now too (_we'll complete them at the UI design phase and when the official rules will be done?_).

The rules are a dialog box that displays scrollable text divided into small sections (chapters), all the text is stored in an object class `RulesObject`. To toggle off the rules, just click outside of them.

I'm open to every comment!

~ v:)


Kaz-ookid (changes requested):
Code looks clean and well encapsulated. Very modular, nice work! View of the activity looks really nice already. Creation of a class for Rules is a great idea. However as code climate highlighted, line 10/11/12 of **RulesObject** could be refactored in a function, as we may end up with more than 3 chapters! 

Nice work!

vrooomer:
@Kaz-ookid Thanks for the review!

In fact, these lines in `RulesObject` are only a mock version now, and I just duplicated the lines to create a bunch of empty chapters. When we'll write the correct rules, no function will be necessary as `RulesChapter` takes a `title` and a `content` as arguments. We won't need more thing to create texts.

Do you get my point?
If it's okay then let's merge this branch.

~ v:)

PR #16, opened 2023-03-16
Feat/position and distance tracking

AmLibra:
Added a visual representation of the phone's current location, that updates in real time as the phone's localization chances.
Merged the Distance Walked counter into the map view in OSM.
Added a dummy position marker on EPFL's Agora (Donut), as a proof of concept of being able to compute the distance to any Marked location on the map, which will be useful when implementing location related gameplay elements.


jadtl (approved):
Thanks for the PR, looks good to me :)

PR #17, opened 2023-03-17
Add EPFL campus layout

jadtl:
This PR adds a layout of the EPFL campus tile source on top of the Open Street Maps tile source.


tamaraaPrs (approved):
Perfect!

PR #18, opened 2023-03-17
Feature/profile

tamaraaPrs:
Some features are missing on this feature such as the ability to change our appearance in the game or select our trophies. Much of the profile GUI is ready in this pull request though!


tamaraaPrs:
And this PR contains all the previous PR of @Jharaxus

Jharaxus (approved):
This looks good to me!
Maybe some other tests could be added to check the well behavior of the interface (are all achievements displayed for example?) but this can be done in the future when the app gets more integrated with the firebase back-end

app/src/main/java/com/github/polypoly/app/game/Trophy.kt
```diff
@@ -0,0 +1,39 @@
+package com.github.polypoly.app.game
+
+/**
+ * A trophy that can be win during a game or at the end of a game to reward a particular
+ * success of the player.
+ */
+class Trophy(
```
AlSchlo, line 7: FYI this can be a
`data class` in Kotlin. It's the equivalent of a `record` in Java.

PR #21, opened 2023-03-17
Feature/firebase user profile support

Jharaxus:


app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
```diff
@@ -0,0 +1,159 @@
...
+open class RemoteDB(
+    private val db: FirebaseDatabase?
+) : IRemoteStorage {
+
+    private lateinit var rootRef: DatabaseReference
+    private lateinit var usersProfilesRef: DatabaseReference
+
+    init {
+        if (db != null) {
+            rootRef = db.getReference()
```
Kaz-ookid, line 26: kotlin allows to use db.reference directly

Kaz-ookid:
In the DB you've created a 'users' list, where users are objects that contain their id. But you also created a list 'all_ids' (see screen). I figured out that you did this to be able to be able to check wether a user already is registered in our DB, for example in **addUser** at line 81: `if (ids.contains(userId))`. 

But do not agree with this structure, because we then have two places to update whenever we remove/add a user or do anything related to user id's. I think that we should have only one list, for example **users**, where keys would be user id's and value would be the user object. That way we can merge both lists. 

What do you think? :)

Screen : 
image

Jharaxus:
> In the DB you've created a 'users' list, where users are objects that contain their id. But you also created a list 'all_ids' (see screen). I figured out that you did this to be able to be able to check wether a user already is registered in our DB, for example in **addUser** at line 81: `if (ids.contains(userId))`.
> 
> But do not agree with this structure, because we then have two places to update whenever we remove/add a user or do anything related to user id's. I think that we should have only one list, for example **users**, where keys would be user id's and value would be the user object. That way we can merge both lists.
> 
> What do you think? :)
> 
> Screen : image

This indeed makes a lot of sense, keeping consistency across several locations is a pain so I will simplify it !

app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
Jharaxus: This is now needed as we specify a root path to differentiate between the "live" and "test" parts of the db

app/src/main/java/com/github/polypoly/app/MapActivity.kt
```diff
@@ -59,6 +63,12 @@ class MapActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
+        val db = Firebase.database
+        try {
+            db.setPersistenceEnabled(false)
+        } catch(_: java.lang.Exception) { }
+        GlobalInstances.remoteDB = RemoteDB(db, "live")
+
```
AmLibra, line 71: Not sure to understand what this is doing here. API specific code that isn't too self explicit (i.e I am assuming this initializes the database or something but it might be more fitting to encapsulate it in a function called "initDB" for example): db.setPersistence has no meaning to someone who has not read FireBase Doc or who has no DB knowledge maybe!

app/src/main/java/com/github/polypoly/app/game/User.kt
```diff
@@ -1,30 +1,47 @@
...
     override fun toString(): String {
         return "User{$id: $name}"
     }
+
+    override fun equals(other: Any?): Boolean {
+        return other is User &&
+                id == other.id && name == other.name && bio == other.bio &&
+                skin == other.skin && stats == other.stats
+    }
 }
```
AmLibra, line 47: Maybe consider documenting even stub classes: User is documented but not Skin or Stats.

app/src/main/java/com/github/polypoly/app/global/GlobalInstances.kt
```diff
@@ -0,0 +1,22 @@
+package com.github.polypoly.app.global
+
+import com.github.polypoly.app.network.RemoteDB
+import java.util.concurrent.atomic.AtomicBoolean
+
```
AmLibra, line 5: Documentation lacking

app/src/main/java/com/github/polypoly/app/global/Settings.kt
```diff
@@ -0,0 +1,17 @@
+package com.github.polypoly.app.global
+
```
AmLibra, line 2: Documentation lacking

app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
```diff
@@ -0,0 +1,128 @@
+package com.github.polypoly.app.network
+
+import com.github.polypoly.app.game.User
+import com.github.polypoly.app.global.Settings.Companion.DB_USERS_PROFILES_PATH
+import com.google.firebase.database.DatabaseReference
+import com.google.firebase.database.FirebaseDatabase
+import com.google.firebase.database.ktx.getValue
+import java.util.concurrent.CompletableFuture
+
```
AmLibra, line 9: Missing documentation (not talking about private and/or overridden functions)

AmLibra (approved):
Overall good and maintainable code. Documentation is lacking but it is expectable this early in the process. Tests are good enough.

app/src/main/java/com/github/polypoly/app/MapActivity.kt
Jharaxus: It provides better consistency between the local firebase database cache and the remote one. I added a comment to make it clearer for now, this might actually become useless if we manually cache data in a further sprint.

app/src/main/java/com/github/polypoly/app/game/User.kt
Jharaxus: Done 👍

app/src/main/java/com/github/polypoly/app/global/GlobalInstances.kt
Jharaxus: Done 👍

app/src/main/java/com/github/polypoly/app/global/Settings.kt
Jharaxus: Done 👍

app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
Jharaxus: Done 👍

app/src/androidTest/java/com/github/polypoly/app/network/RemoteDBTest.kt
```diff
@@ -0,0 +1,153 @@
...
+        addUsersToDB(listOf(testUser1))
+
+        assertTrue(remoteDB.updateUser(userUpdated).get(TIMEOUT_DURATION, TimeUnit.SECONDS))
+
+        val userUpdatedFound = remoteDB.getUserWithId(userUpdated.id).get(TIMEOUT_DURATION, TimeUnit.SECONDS)
+        assertEquals(userUpdated, userUpdatedFound)
+    }
+
+    @Test
+    fun unergisteredUserCannotBeUpdated() {
```
AlSchlo, line 144: Fix typo

app/src/main/java/com/github/polypoly/app/game/User.kt
```diff
@@ -1,30 +1,55 @@
...
+/**
+ * All stats concerning a given player.
+ * This will be further developed in the future.
+ */
+class Stats constructor(
+    val accountCreation: Int = 0, //> TODO improve unix-representation to something better
+    val lastConnection: Int = 0,
+    val numberOfWins: Int = 0
+) {
+    override fun equals(other: Any?): Boolean {
```
AlSchlo, line 29: 😨

app/src/main/java/com/github/polypoly/app/game/User.kt
```diff
@@ -1,30 +1,55 @@
...
-    val skin: Skin,
-    val stats: Stats
+class User constructor(
+    val id: Long = 0,
+    val name: String = "no-name",
+    val bio: String = "no-bio",
+    val skin: Skin = Skin(),
+    val stats: Stats = Stats()
 ){
     override fun toString(): String {
```
AlSchlo, line 46: Your `toString` will automatically be nicely defined in a `data class`.

app/src/main/java/com/github/polypoly/app/game/User.kt
```diff
@@ -1,30 +1,55 @@
...
         return "User{$id: $name}"
     }
+
+    override fun equals(other: Any?): Boolean {
```
AlSchlo, line 50: 😱

app/src/main/java/com/github/polypoly/app/global/GlobalInstances.kt
```diff
@@ -0,0 +1,27 @@
...
+        var remoteDB: RemoteDB = RemoteDB.InvalidRemoteDB
+            get() {
+                if (!isRemoteDBSet.get()) {
+                    throw java.lang.IllegalStateException("RemoteDB is not yet setup")
+                }
+                return field
+            }
+            set(value) {
+                if (isRemoteDBSet.compareAndSet(false, true)) {
+                    field = value // cannot throw an IllegualArgumentException because it is not yet determinstic how the order of instantiation works for static methods
```
AlSchlo, line 23: Typo

app/src/main/java/com/github/polypoly/app/global/GlobalInstances.kt
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Instances shared across the entire app for consistency purpose. This consist in:
+ * - a RemoteDB instance which is our connection to the firebase DB
+ * - more to be added later
+ */
+class GlobalInstances {
+    companion object {
+        private val isRemoteDBSet: AtomicBoolean = AtomicBoolean(false)
+        var remoteDB: RemoteDB = RemoteDB.InvalidRemoteDB
```
AlSchlo, line 14: Could use `lazy var`

app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
```diff
@@ -0,0 +1,131 @@
...
+            registerPromise
+        }
+    }
+
+    override fun updateUser(user: User): CompletableFuture<Boolean> {
+        return getAllUsersIds().thenCompose { allIds ->
+            val registerPromise = CompletableFuture<Boolean>()
+
+            if (!allIds.contains(user.id)) {
```
AlSchlo, line 91: This is O(N) in number of users! What will happen if 1000 people play Polypoly?
This is a major design flaw: you should fetch the user directly with its id.

app/src/main/java/com/github/polypoly/app/MapActivity.kt
```diff
@@ -59,6 +63,12 @@ class MapActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
+        val db = Firebase.database
+        try {
+            db.setPersistenceEnabled(false) // disable local caching and do only write-through instead
+        } catch(_: java.lang.Exception) { }
```
SolalPirelli, line 69: Ignoring all exceptions deserves a comment :)

app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
```diff
@@ -0,0 +1,131 @@
...
+
+            registerPromise
+        }
+    }
+
+    fun getUnderlyingDB(): FirebaseDatabase{
+        return db!!
+    }
+
+    companion object InvalidRemoteDB: RemoteDB(null, "") {
```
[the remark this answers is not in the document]
SolalPirelli: FWIW, there is such a thing as the [null object pattern](https://en.wikipedia.org/wiki/Null_object_pattern), however this is not it: there's no point in throwing those exceptions compared to literally just using `null`. It would make sense if you intended to, e.g., return no users at all to represent an empty/unavailable DB.

app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
```diff
@@ -0,0 +1,131 @@
...
+        override fun getUserWithId(userId: Long): CompletableFuture<User> {
+            throw IllegalAccessError("This RemoteDB is invalid")
```
SolalPirelli, line 112: This is not the right kind of exception to throw. `Error`s should basically never be thrown by your own code, only by the runtime, they indicate something is very very wrong. In this case `IllegalAccessError` is intended to mean "you're trying to access a field you don't have access to because you're loading incompatible JARs".

PR #23, opened 2023-03-17
Refactoring + Merging features from sprint 1

vrooomer:
This big PR comes from the branch that we used for the demo, that merged all the work in a single activity: `WelcomeActivity`.
We decided to delete `MainActivity` and `GreetingActivity` as they won't be useful anymore.

Here is how the buttons in the new "main" activity (`WelcomeActivity`) work so far, we can discuss it later:
- Join Game: join a game with group id
- Create Game: nothing

At the bottom, from left to right:
- Button 1: rules
- Button 2: map
- Button 3: profile
- Button 4: settings

Also, to clean up our files, I created different packages: All the classes that concern the menu of the game are now in `com.github.polypoly.app.menu` and all the non-activity classes related to the menu are in `com.github.polypoly.app.menu.kotlin`

===========

Finally, while checking the tests, I saw that `ProfileActivity` is barely tested (check `WelcomeActivityTest` for inspiration). Same for the activities related to the map and the DB back-end (let me know what parts are untestable)

As this PR is quite big, I would like to get more than one review before we change to this version!

~ v:)


app/src/androidTest/java/com/github/polypoly/app/WelcomeActivityTest.kt
```diff
@@ -17,7 +21,7 @@ class WelcomeActivityTest {
     private val optionButtonRules = composeTestRule.onNodeWithContentDescription("Show Rules")
     private val optionButton2 = composeTestRule.onNodeWithContentDescription("optionButton2")
```
Jharaxus, line 22: UI component should have a more meaningful name and id to express their intent

Jharaxus: This works also for optionButton3

Jharaxus: Why are they call **Option**Button btw?

app/src/main/java/com/github/polypoly/app/menu/JoinGroupActivity.kt
```diff
@@ -155,8 +156,8 @@ class JoinGroupActivity : ComponentActivity() {
      * This function launches the group room activity and passes the group code to it.
      */
     private fun joinGroupRoom(mContext : Context) {
-        val groupIntent = Intent(mContext, GreetingActivity::class.java)
-        groupIntent.putExtra("groupCode", groupCode)
+        //val groupIntent = Intent(mContext, GreetingActivity::class.java)
+        //groupIntent.putExtra("groupCode", groupCode)
```
Jharaxus, line 161: The commented code should be removed from PR

app/src/main/java/com/github/polypoly/app/menu/kotlin/GameMusic.kt
```diff
@@ -0,0 +1,28 @@
...
+    fun setVolume(value: Float) {
+        var validValue = max(value, 1f)
+        validValue = min(validValue, 0f)
+        mediaPlayer.setVolume(validValue, validValue)
+    }
+
+    fun mute() {
+        setVolume(0f)
+    }
+
```
AmLibra, line 26: Please avoid useless whitespaces.

app/src/main/java/com/github/polypoly/app/menu/kotlin/GameMusic.kt
```diff
@@ -0,0 +1,28 @@
...
+    private lateinit var mediaPlayer: MediaPlayer
+
+    fun startSong() {
+        mediaPlayer = MediaPlayer.create(context, songId)
+        mediaPlayer.isLooping = true
+        mediaPlayer.start()
+    }
+
+    fun setVolume(value: Float) {
+        var validValue = max(value, 1f)
```
AmLibra, line 18: Avoid Magic numbers!

app/src/main/java/com/github/polypoly/app/menu/WelcomeActivity.kt
```diff
@@ -160,7 +173,9 @@ class WelcomeActivity : ComponentActivity() {
             ) {
                 Surface(
                     color = MaterialTheme.colors.primary,
-                    modifier = Modifier.fillMaxWidth(0.95f).fillMaxHeight(0.95f)
+                    modifier = Modifier
+                        .fillMaxWidth(0.95f)
+                        .fillMaxHeight(0.95f)
```
Jharaxus, line 178: Magic numbers should be avoided

app/src/main/java/com/github/polypoly/app/menu/WelcomeActivity.kt
```diff
@@ -160,7 +173,9 @@ class WelcomeActivity : ComponentActivity() {
             ) {
                 Surface(
                     color = MaterialTheme.colors.primary,
-                    modifier = Modifier.fillMaxWidth(0.95f).fillMaxHeight(0.95f)
+                    modifier = Modifier
```
AmLibra, line 176: Avoid Magic numbers!

app/src/main/java/com/github/polypoly/app/menu/kotlin/GameMusic.kt
```diff
@@ -0,0 +1,28 @@
...
+        validValue = min(validValue, 0f)
```
Jharaxus, line 19: You can use `clamp` to bind validValue between 0f ans 1f

Jharaxus (changes requested):
Some minor things should be changed, but the overall merge is good :)

app/src/main/java/com/github/polypoly/app/menu/ProfileModifyingActivity.kt
```diff
@@ -126,6 +126,7 @@ class ProfileModifyingActivity : ComponentActivity() {
 
     /**
      * The field where the player can write his/her description
```
AmLibra, line 128: Documentation like this is very good, less pointless // prepended comments would be better !

AmLibra (changes requested):
This seems to be a good agglomeration of everyone's work, but could use a bit of refactoring and rigor.

app/src/androidTest/java/com/github/polypoly/app/WelcomeActivityTest.kt
vrooomer: I agree with you, the problem is that they don't have a fixed utility so far, I put some launching intents but only for the demo.
We need to decide what button does what, I'm gonna change that

app/src/main/java/com/github/polypoly/app/menu/kotlin/GameMusic.kt
vrooomer: Nice! I didn't know about this, thanks :)

app/src/main/java/com/github/polypoly/app/menu/WelcomeActivity.kt
vrooomer: I also agree with that for magic numbers. But in this case these are just parameters used locally, I don't think using a constant there should be a good idea as it depends on each composable.

vrooomer:
I fixed most of the issues. I just disagree with the "magic number" issue, as it is only a parameter given to a unique composable and not global at all.

~ v:)

AmLibra (approved):

app/src/androidTest/java/com/github/polypoly/app/WelcomeActivityTest.kt
Jharaxus: Great, this looks fine now that you added details regarding what option they refer to

app/src/main/java/com/github/polypoly/app/menu/WelcomeActivity.kt
Jharaxus: Ok we discussed this together and I aggree with your point saying that it is a more global issue than just this PR, therefore I created a specific [issue](https://github.com/polypoly-team/android-app/issues/40) that we can address in a later sprint.

app/src/main/java/com/github/polypoly/app/menu/kotlin/GameMusic.kt
Jharaxus: No problem :)

Jharaxus (approved):

PR #41, opened 2023-03-22
change the color theme

tamaraaPrs:
Just a little PR to change the color theme and make the new color palette available

### DIFF
--- a/app/src/main/java/com/github/polypoly/app/menu/ProfileActivity.kt
+++ b/app/src/main/java/com/github/polypoly/app/menu/ProfileActivity.kt
@@ -20,7 +20,9 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
+import androidx.compose.ui.draw.shadow
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.modifier.modifierLocalConsumer
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.text.style.TextAlign
@@ -29,6 +31,8 @@ import androidx.compose.ui.unit.dp
 import com.github.polypoly.app.game.allTrophies
 import com.github.polypoly.app.network.FakeRemoteStorage
 import com.github.polypoly.app.ui.theme.PolypolyTheme
+import com.github.polypoly.app.ui.theme.purpleVeryDark
+import com.github.polypoly.app.ui.theme.purpleVeryLight
 
 class ProfileActivity : ComponentActivity() {
 
@@ -63,7 +67,10 @@ class ProfileActivity : ComponentActivity() {
             verticalArrangement = Arrangement.Top,
             horizontalAlignment = Alignment.CenterHorizontally
         ) {
-            Surface(elevation = 8.dp) {
+            Surface(
+                elevation = 8.dp,
+                color = MaterialTheme.colors.background
+            ) {
                 Profile()
             }
             Statistics()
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/polypoly/app/menu/ProfileActivity.kt
```diff
@@ -79,7 +86,8 @@ class ProfileActivity : ComponentActivity() {
         val mContext = LocalContext.current
         Column(
             modifier = Modifier
-                .padding(all = 30.dp),
+                .padding(all = 30.dp)
+                .background(color = Color.White),
```
vrooomer, line 90: You should avoid things like that, you created a palette so it's better to use it. What if we would like to change our background color everywhere. You could use `MaterialTheme.colors.background`

app/src/main/java/com/github/polypoly/app/menu/ProfileActivity.kt
```diff
@@ -93,15 +101,23 @@ class ProfileActivity : ComponentActivity() {
...
                 onClick = {
                     val profileModifyingIntent = Intent(mContext, ProfileModifyingActivity::class.java)
                     profileModifyingIntent.putExtra("userId", userId)
                     startActivity(profileModifyingIntent)
                 },
                 shape = CircleShape,
-                modifier = Modifier.testTag("modifyProfileButton"),
+                modifier = Modifier
+                    .testTag("modifyProfileButton")
+                    .background(color = Color.White),
```
vrooomer, line 117: Same issue here

app/src/main/java/com/github/polypoly/app/menu/ProfileActivity.kt
```diff
@@ -93,15 +101,23 @@ class ProfileActivity : ComponentActivity() {
...
                 colors = ButtonDefaults.buttonColors(
-                    backgroundColor = Color(red = 240, blue = 240, green = 240))
+                    backgroundColor = purpleVeryLight,
+                    disabledBackgroundColor = Color.White)
```
vrooomer, line 120: And here

app/src/main/java/com/github/polypoly/app/menu/ProfileActivity.kt
```diff
@@ -168,7 +184,7 @@ class ProfileActivity : ComponentActivity() {
         Box(
             modifier = Modifier
                 .clip(CircleShape)
-                .background(color = if (won) MaterialTheme.colors.primary else Color.DarkGray)
+                .background(color = if (won) MaterialTheme.colors.secondary else purpleVeryDark)
```
vrooomer, line 187: Here too, even if you use the colors defined in `PolypolyTheme`, I think it's better if we really only stick to the palettes and only the palettes.

vrooomer (changes requested):
The overall code looks great and I have not a lot to say about it as the PR is relatively small.
However I found some issues that I quote below.

Good job on the colors, they're great! 
~ v:)

app/src/main/java/com/github/polypoly/app/menu/ProfileActivity.kt
```diff
@@ -93,15 +101,23 @@ class ProfileActivity : ComponentActivity() {
...
+                ),
                 onClick = {
                     val profileModifyingIntent = Intent(mContext, ProfileModifyingActivity::class.java)
                     profileModifyingIntent.putExtra("userId", userId)
                     startActivity(profileModifyingIntent)
                 },
                 shape = CircleShape,
-                modifier = Modifier.testTag("modifyProfileButton"),
+                modifier = Modifier
+                    .testTag("modifyProfileButton")
```
vrooomer, line 116: Also, use the `testTag` only when it is really the only option to test a composable. Here it is not the case as you could use `composeTestRule.onNodeWithText("Modify profile")`.

vrooomer: Okay, I agree that these strings we've been using for UI so far is a bad habit. I'll create an issue to create an external file with all the games strings.

vrooomer (approved):
LGTM, thanks for the changes!

~ v:)

PR #44, opened 2023-03-23
Fix CampusTileSource

jadtl:
A small fix to the campus tile source so that it picks a random server at every request, the bound for the random number is also fixed.

### DIFF
--- a/app/src/main/java/com/github/polypoly/app/MapActivity.kt
+++ b/app/src/main/java/com/github/polypoly/app/MapActivity.kt
@@ -82,7 +82,7 @@ class MapActivity : ComponentActivity() {
 
             addMarkerTo(mapView, donutPosition, "Donut")
 
-            val campusTileSource = CampusTileSource(Random.nextInt(2), 0)
+            val campusTileSource = CampusTileSource(0)
             val tileProvider = MapTileProviderBasic(applicationContext, campusTileSource)
             val tilesOverlay = TilesOverlay(tileProvider, applicationContext)
             mapView.overlays.add(tilesOverlay)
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/github/polypoly/app/MapActivity.kt
```diff
@@ -216,9 +216,9 @@ class MapActivity : ComponentActivity() {
     }
 
 
-    class CampusTileSource(private val serverId: Int, private val floorId: Int) : OnlineTileSourceBase("EPFLCampusTileSource", 0, 18, 256, ".png", arrayOf()) {
+    class CampusTileSource(private val floorId: Int) : OnlineTileSourceBase("EPFLCampusTileSource", 0, 18, 256, ".png", arrayOf()) {
         override fun getTileURLString(pMapTileIndex: Long): String {
-            return "https://plan-epfl-tiles$serverId.epfl.ch/1.0.0/batiments/default/20160712/$floorId/3857/${MapTileIndex.getZoom(pMapTileIndex)}/${MapTileIndex.getY(pMapTileIndex)}/${MapTileIndex.getX(pMapTileIndex)}.png"
+            return "https://plan-epfl-tiles${Random.nextInt(3)}.epfl.ch/1.0.0/batiments/default/20160712/$floorId/3857/${MapTileIndex.getZoom(pMapTileIndex)}/${MapTileIndex.getY(pMapTileIndex)}/${MapTileIndex.getX(pMapTileIndex)}.png"
```
vrooomer, line 221: Maybe putting a 3 here is a bad thing as it creates a magic number?

vrooomer:
Perfect, thanks! I found a small issue, I let you check this before the merge.

~ v:)

vrooomer (changes requested):
Perfect, thanks! I found a small issue, I let you check this before the merge.

~ v:)

app/src/main/java/com/github/polypoly/app/MapActivity.kt
jadtl: You're right, and it's not clear why it is 3, I'll change that, thanks.

jadtl:
Is it better like this?

vrooomer (approved):
It doesn't fix it but I guess for the usage of it, it is acceptable :)

PR #45, opened 2023-03-23
Feat/building localizations

AmLibra:
Merged and refactored my code with Jad's. One notable change is using an viewModel to track the walked distance to better encapsulate what is Jetpack Compose UI and what is not.
Added more building locations, a concept of zone encapsulating a set of locations with the same colour.
Also allowing for easy add remove and edit of locations and zones.
We can now also modify easily the icon of the pins and recolor them any colour desired (taken from the zone of the location).
This is draft as I am not done cleaning up code, commenting and testing.


app/src/main/java/com/github/polypoly/app/game/Localization.kt
```diff
@@ -1,20 +1,44 @@
...
+/**
+    An enum class representing the different levels of a [Localization].
+    @property EMPTY the lowest level of a [Localization], when it is not yet owned.
+    @property STUDY_ROOM a level of a [Localization], when it has been acquired by the player.
+    @property AMPHITHEATER a level of a [Localization], when it has been upgraded once.
+    @property ARENA a level of a [Localization], when it has been upgraded twice.
+    @property STADIUM the highest level of a [Localization], when it has been upgraded thrice.
+ */
 enum class LocalizationLevel {
     EMPTY, STUDY_ROOM, AMPHITHEATER, ARENA, STADIUM
```
tamaraaPrs, line 25: I like the idea very much, but I find that Arena and Stadium are no longer related to EPFL and I think we can do better. I propose instead:
- EMPTY
- STUDY_ROOM
- AMPHUTHEATER
- FLOOR
- BUILDING

app/src/main/java/com/github/polypoly/app/map/LocalizationRepository.kt
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Repository for providing the zones and localizations on the map.
+ */
+object LocalizationRepository {
+
+    /**
+     * List of localizations in the EPFL campus area.
+     */
+    private val localizations = listOf(
```
tamaraaPrs, line 16: If I understand correctly, this class is used to represent the buildings in general in a game. I think that it could be good to separate this class in two classes because otherwise it could complicate things because here we mix two different concepts:
- The buildings available in general in all the games. That is to say an exhaustive list of the buildings of the EPFL that will never change. 
- The current state of the buildings in the game (the current level, how much is the tax now, etc)

I propose two classes to represent these two concepts:
- `object LocalisationRepository` contains all the Localization possible in the game, with their name, their zone, their position, and evently a price table but I think it's better to calculate it according to the zone.
- `class LocalisationState` contains a list with the references of localisations of `LocalisationRepository`, their current owners, their current levels, and a method to calculate the tax based on the price and the level of the localisations. We can give an instance of this class as an attribute of a game.

app/src/main/java/com/github/polypoly/app/map/LocalizationRepository.kt
```diff
@@ -0,0 +1,94 @@
...
+        Localization(
+            name = "BC",
+            basePrice = 200,
+            baseTaxPrice = 20,
+            baseMortgagePrice = 50,
+            currentLevel = LocalizationLevel.STUDY_ROOM,
```
tamaraaPrs, line 22: All localizations are initialized to EMPTY at the beginning

app/src/main/java/com/github/polypoly/app/map/LocalizationRepository.kt
```diff
@@ -0,0 +1,94 @@
...
+object LocalizationRepository {
+
+    /**
+     * List of localizations in the EPFL campus area.
+     */
+    private val localizations = listOf(
+        Localization(
+            name = "BC",
+            basePrice = 200,
+            baseTaxPrice = 20,
```
tamaraaPrs, line 20: The tax price can be computed with the basePrice and the level for example, It makes sure that all taxes are fair

tamaraaPrs (changes requested):
It is very well commented and the code is very clean. I just think that the way the localizations are represented in the code should be changed, as I explained in the comment.

app/src/main/java/com/github/polypoly/app/game/Localization.kt
AmLibra: While I totally agree, I think Floor could be taken out of context and would prefer replacing FLOOR by BUILDING directly, and BUILDING could then become LEARNING CENTER or similar (as it does not make much to go from amphitheater to a floor, a floor would already be existing already in order for the amphitheater to be there supposedly). Another suggestion would be Library or Laboratory

app/src/main/java/com/github/polypoly/app/map/LocalizationRepository.kt
AmLibra: As you are pointing out, it is indeed better for these two concepts to be separated, I will simply remove the currentLevel attribute as it seems to be the only one that would better fit LocalizationState, the rest of the current attributes seem to be intrinsically linked to the localization itself and should not change during the game.
I'm down to work on LocalizationState next week though !

app/src/main/java/com/github/polypoly/app/map/LocalizationRepository.kt
AmLibra: As stated above this is all copilot stuff I did not even look at twice, so it will be leaving as discussed in the above comment anyways.

app/src/main/java/com/github/polypoly/app/map/LocalizationRepository.kt
AmLibra: Ditto

app/src/main/java/com/github/polypoly/app/game/Localization.kt
Jharaxus: I would go for LEVEL_0, LEVEL_1, LEVEL_2, LEVEL_3 and LEVEL_MAX as the actual meaning in game-terms should be taken care of by the front-end. This would also let us handle multiple languages and concentrate on the inner meaning of this enum (it represents the possible levels of a localization), and not how we interpret them (is level 1 a study room or an exercise room in our game rp? We don't care, it is the second level of evolution)

AmLibra: Sounds good to me

app/src/main/java/com/github/polypoly/app/game/Localization.kt
```diff
@@ -15,14 +15,9 @@ class Zone(
...
     An enum class representing the different levels of a [Localization].
-    @property EMPTY the lowest level of a [Localization], when it is not yet owned.
-    @property STUDY_ROOM a level of a [Localization], when it has been acquired by the player.
-    @property AMPHITHEATER a level of a [Localization], when it has been upgraded once.
-    @property ARENA a level of a [Localization], when it has been upgraded twice.
-    @property STADIUM the highest level of a [Localization], when it has been upgraded thrice.
  */
 enum class LocalizationLevel {
-    EMPTY, STUDY_ROOM, AMPHITHEATER, ARENA, STADIUM
+    LEVEL_0, LEVEL_1, LEVEL_2, LEVEL_3, LEVEL_MAX
```
tamaraaPrs, line 20: We can use that for now and we'll see about the names later.

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
tamaraaPrs, line 1: Don't forget to comment your code!

tamaraaPrs (approved):
This is great, just don't forget to add comments but overall very good job!

app/src/androidTest/java/com/github/polypoly/app/map/LocalizationRepositoryTest.kt
```diff
@@ -0,0 +1,17 @@
+package com.github.polypoly.app.map
+
+import org.junit.Test
+
+// These tests are not really useful, but they're here symbolically, this is only a data class
```
AlSchlo, line 5: Why is it in /AndroidTest then?

app/src/androidTest/java/com/github/polypoly/app/map/LocalizationRepositoryTest.kt
```diff
@@ -0,0 +1,17 @@
...
+class LocalizationRepositoryTest {
+    @Test
+    fun testGetLocalizationsWorks() {
+        assert(LocalizationRepository.getZones().flatMap { zone -> zone.localizations }
```
AlSchlo, line 9: You should not use `assert` in tests. You can use hamcrest's `isTrue` for this.

app/src/androidTest/java/com/github/polypoly/app/map/MapActivityTest.kt
```diff
@@ -0,0 +1,98 @@
...
+
+@RunWith(AndroidJUnit4::class)
+class MapActivityTest {
+
+    @get:Rule
+    val composeTestRule = createAndroidComposeRule<MapActivity>()
+
+    @Before
+    fun setUp() {
+        runBlocking { delay(5000) } // TODO: Find a better way to wait for the UI to update
```
AlSchlo, line 24: You can code a `TryOnRepeat` function for that (that checks whether the UI (or other predicate) is satisfied. Also, waiting 5 seconds is super long 😱 !

app/src/androidTest/java/com/github/polypoly/app/map/MapActivityTest.kt
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    @Test
+    fun mapActivity_UIComponents_Displayed() {
```
AlSchlo, line 28: Weird naming convention

app/src/androidTest/java/com/github/polypoly/app/map/MapActivityTest.kt
```diff
@@ -0,0 +1,98 @@
...
+    @Test
+    fun mapActivity_ResetButton_Clicked_DistanceReset() {
+        fun formattedDistance(distance: Float): String {
+            return if (distance < 1000) "${"%.1f".format(distance)}m"
+            else "${"%.1f".format(distance / 1000)}km"
+        }
+
+        composeTestRule.onNodeWithTag("resetButton").performClick()
+
+        runBlocking { delay(500) }
```
AlSchlo, line 43: 😨

PR #46, opened 2023-03-23
Add heads-up display

jadtl:
This PR adds a heads-up display to the map activity.

It displays:

- Player information like their balance.
- A drop down menu for information about the game and the other players.


app/src/main/java/com/github/polypoly/app/MapActivity.kt
```diff
@@ -154,6 +161,48 @@ class MapActivity : ComponentActivity() {
         }
     }
 
+    @Composable
+    fun Hud(data: PlayerGlobalData, round: Int) {
```
Kaz-ookid, line 165: At a final stage of this feature, you should use current authenticated better than `data` argument. You could use `FirebaseAuth.getInstance().currentUser?.let{ //something }` (requires to add some dependencies), 
from which you can get the uid -> find the user data in data-base once it is all setup in Firebase-user-support PR.

app/src/main/java/com/github/polypoly/app/MapActivity.kt
```diff
@@ -154,6 +161,48 @@ class MapActivity : ComponentActivity() {
...
+        Box(modifier = Modifier.fillMaxWidth()) {
+            Box(
+                modifier = Modifier
+                    .padding(16.dp)
+                .background(Color.White)
+                .border(1.dp, Color.Black)
+                .align(Alignment.TopStart)
+            ) {
+                Row(modifier = Modifier.padding(8.dp)) {
```
Kaz-ookid, line 174: Padding of **8.dp** looks redundant. We could create a class filled with UI constants, such as a universal padding :)!

app/src/main/java/com/github/polypoly/app/MapActivity.kt
Kaz-ookid, line 1: HUD looks great! nice work on the look

app/src/main/java/com/github/polypoly/app/MapActivity.kt
```diff
@@ -154,6 +161,48 @@ class MapActivity : ComponentActivity() {
...
+                    Button(
+                        onClick = {},
+                        modifier = Modifier
+                            .size(70.dp)
+                            .semantics { contentDescription = "" },
+                        shape = RoundedCornerShape(10.dp),
```
tamaraaPrs, line 180: In the graphic chart we use round buttons when they are small. But it's important to ask yourself what this button is for, because it might be more relevant to make it look different from the graphic chart and to put it bigger, and here with your code I'm not sure I understand what it is for.

app/src/main/java/com/github/polypoly/app/MapActivity.kt
tamaraaPrs: Take care to respect the graphic charter and to use `MaterialTheme.colors`. Also pay attention to the modularity.

tamaraaPrs (changes requested):

tamaraaPrs (changes requested):
Don't forget to put comments to explain in detail each element do

app/src/main/java/com/github/polypoly/app/MapActivity.kt
```diff
@@ -154,6 +161,48 @@ class MapActivity : ComponentActivity() {
         }
     }
 
+    @Composable
+    fun Hud(data: PlayerGlobalData, round: Int) {
+        Box(modifier = Modifier.fillMaxWidth()) {
```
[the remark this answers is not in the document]
jadtl: If I remove the first box, then the `align` function for the second box becomes unavailable, and I can't place it on the top left, I can't find how to do this differently.

app/src/main/java/com/github/polypoly/app/MapActivity.kt
jadtl: It's a button that displays the full game information, so I think it should be the same size as the buttons, what do you think? I've made this clearer in the code.

app/src/main/java/com/github/polypoly/app/MapActivity.kt
```diff
@@ -154,6 +161,48 @@ class MapActivity : ComponentActivity() {
...
+            Box(
+                modifier = Modifier
+                    .padding(16.dp)
+                .background(Color.White)
+                .border(1.dp, Color.Black)
```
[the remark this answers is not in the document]
jadtl: Okay, somehow I can't add elevation to the `Box` even though I imported `androidx.compose.material`, I've removed the border and left the box without highlight.

app/src/main/java/com/github/polypoly/app/MapActivity.kt
jadtl: I created a `Padding` class in a package called `utils` to store the constants!

app/src/main/java/com/github/polypoly/app/MapActivity.kt
```diff
@@ -154,6 +161,48 @@ class MapActivity : ComponentActivity() {
...
+                        contentPadding = PaddingValues(0.dp)
+                    ) {
+                        Image(
+                            painter = painterResource(R.drawable.tmp_happysmile),
+                            contentDescription = "",
+                            modifier = Modifier.size(50.dp)
+                        )
+                    }
+                    Column(modifier = Modifier.padding(8.dp)) {
+                        Text(
```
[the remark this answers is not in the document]
jadtl: I've modularized the `Text` elements and the `Button`, is it good like that?

app/src/main/java/com/github/polypoly/app/utils/Padding.kt
```diff
@@ -0,0 +1,11 @@
+package com.github.polypoly.app.utils
+
+import androidx.compose.ui.unit.dp
+
+class Padding {
```
AlSchlo, line 5: Why create a class with nothing in it except a companion object ❓ . Juste create an object directely.

jadtl: Nice catch, thanks!

tamaraaPrs (approved):

PR #51, opened 2023-03-29
Feat join group from list

Kaz-ookid:
Implemented a button in **JoinGroupActivity** that opens a dialog filled with the list of Public and not full groups in the DB. This list if refreshed every 5 seconds. 

It is safe to try and join a group that shows as not full in the client side but is in the DB side, it will just prompt that the group is full and will disappear within the next refresh.

The list of groups is showed as bars that containt the name of the group and the number of players in it, as well as the maximum number of players. It show as so: `Group name         'avatar icon' 3/5`
These bars are clickable, and once clicked un-rolls the whole details of the lobby: name and profile pictures of the players in the lobby, game mode, round duration and a `Join` button that lets the user join the given group.

Tests are not yet implemented and that is the reason this is a **draft pull request**

Graphic chart is not yet respected


Kaz-ookid:
Replaced mock implementation with the use of Fake DB that uses the same interface as the real DB. Will work well once the real DB queries are up. Tests are not yet finished but should work well anyway.

app/src/main/java/com/github/polypoly/app/network/FakeRemoteStorage.kt
```diff
@@ -22,28 +82,65 @@ class FakeRemoteStorage : IRemoteStorage {
...
+        gameLobbyJoinable1.addUser(testUser2)
+        gameLobbyJoinable1.addUser(testUser3)
+
+        gameLobbyJoinable2.addUser(testUser2)
+        gameLobbyJoinable2.addUser(testUser3)
+        gameLobbyJoinable2.addUser(testUser4)
+
+        gameLobbyPrivate.addUser(testUser2)
+        gameLobbyPrivate.addUser(testUser3)
+        gameLobbyPrivate.addUser(testUser4)
```
Jharaxus, line 100: This could be refactored with a function `addUsersToGroups` to make one-line assignments instead per `GameLobby` instead.

Jharaxus (changes requested):
A couple of refactoring is still needed, but great overall job!

Jharaxus (approved):
Great, this looks to work fine !

app/src/androidTest/java/com/github/polypoly/app/JoinGameLobbyActivityTest.kt
```diff
@@ -0,0 +1,91 @@
...
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class JoinGameLobbyActivityTest {
+
+    @get:Rule
+    val composeTestRule = createAndroidComposeRule<JoinGameLobbyActivity>()
+
+    @Test
+    fun launchActivity_componentsDisplayed() {
```
AlSchlo, line 19: Weird mix of camelCase and snake_case

app/src/main/java/com/github/polypoly/app/network/FakeRemoteStorage.kt
```diff
@@ -2,18 +2,65 @@ package com.github.polypoly.app.network
...
 import java.time.LocalDateTime
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.Future
+import kotlin.time.Duration.Companion.hours
 
 /**
- * A fake remote storage to test the functionalities without the database
+ * A fake remote storage to test the functionalities with compatibility for later use of the DB
  */
 class FakeRemoteStorage : IRemoteStorage {
```
AlSchlo, line 14: Why has FakeRemoteStorage become a storage for game lobbies, users, etc.? This looks like bad abstraction.

app/src/main/java/com/github/polypoly/app/network/IRemoteStorage.kt
```diff
@@ -1,7 +1,9 @@
 package com.github.polypoly.app.network
 
+import com.github.polypoly.app.game.GameLobby
 import com.github.polypoly.app.game.User
 import java.util.concurrent.CompletableFuture
+import java.util.concurrent.Future
 
 interface IRemoteStorage {
```
AlSchlo, line 8: Your remote storage is poorly abstracted. It should not be directly responsible for managing GameLobbies

Issue #40, opened 2023-03-19
Refactor magic numbers in Activities


Jharaxus:
Some magic numbers sometimes appear in the codes of Activities, however, such as pointed [here](https://github.com/polypoly-team/android-app/pull/23#discussion_r1140305979), that doesn't make much sense when read out of context (as it just looks like any magic number --> why does it have this value?).

Therefore, it could be a good improve to our codebase to cross-read all activities' code and look:
- Wether some magic values constants could be merged together and put them in separate UI config files
- For the others, find some way to make them clearer such as turning them into constants located in their activity’s companion object.

Kaz-ookid:
we should adress this issue soon, and create a Constants class in utils/

PR #54, opened 2023-03-30
Feature/profile

tamaraaPrs:
**Improved profile, especially for trophies.** In this PR you will find the following features:

In ProfileActivity:
- Add icons to the trophies.
- Can select a trophy to see its description in the profile page, but the description is shown only if the user has won the trophy, else the bubble shows "???". The trophies not selected have less saturation.
- Now, the layout of the trophies is adaptative to the size of the screen (there are less trophies per row if the screen is smaller).
- Add a background map to the profile page and change the layout
- Add the possibility to have less than 3 trophies display for the other players. In this case, the profile show an empty slot instead of the trophy
- Map the stats display to the actual stats of the player (before it was hardcoded number)

In ProfileModifyingActivity:
- Can select at most 3 trophies the trophies displayed to other players
- Can clear the trophies selection
- Put a limit of line returns in the description field

I also modularized my old code and added some utility functions.


tamaraaPrs:
The tests were not already passing on the main commits before my PR. I don't understand why cause all my tests are passing locally. I don't know what's the reason, but it seems to me that the tests weren't running since this commit from @jadtl :
image

Jharaxus:
Capture d’écran 2023-03-31 à 01 36 18
Capture d’écran 2023-03-31 à 01 36 08

The background map scrolls together with the profile boxes in the profile view, but I think it should stay fixed instead.

tamaraaPrs:
@Jharaxus 
It's not a bug, it's intentional. If you like it otherwise I can change it but I had implemented it this way by choice

Jharaxus:
> @Jharaxus It's not a bug, it's intentional. If you like it otherwise I can change it but I had implemented it this way by choice

Ok, then let's keep it as is for now and decide during a later sprint if we prefer having it the other way :)

app/src/androidTest/java/com/github/polypoly/app/game/PendingGameTest.kt
```diff
@@ -137,10 +140,10 @@ class PendingGameTest {
...
             testDuration, emptyList(), testInitialBalance, testName, testCode)
 
-        val u1 = User(42042043, "test_user1", "", emptySkin, zeroStats)
-        val u2 = User(42042044, "test_user2", "", emptySkin, zeroStats)
-        val u3 = User(42042045, "test_user3", "", emptySkin, zeroStats)
-        val u4 = User(42042046, "test_user4", "", emptySkin, zeroStats)
+        val u1 = User(42042043, "test_user1", "", emptySkin, zeroStats, listOf(), mutableListOf())
+        val u2 = User(42042044, "test_user2", "", emptySkin, zeroStats, listOf(), mutableListOf())
+        val u3 = User(42042045, "test_user3", "", emptySkin, zeroStats, listOf(), mutableListOf())
+        val u4 = User(42042046, "test_user4", "", emptySkin, zeroStats, listOf(), mutableListOf())
```
Jharaxus, line 147: This initialization is redundant, so it would be better to add them as class members testUser2, testUser3, testUser4 and testUser5 instead

app/src/androidTest/java/com/github/polypoly/app/menu/ProfileActivityTest.kt
```diff
@@ -0,0 +1,96 @@
...
+        button.performClick()
+        intended(hasComponent(ProfileModifyingActivity::class.java.name))
+
+        Intents.release()
+    }
+
+    @Test
+    fun canSeeTheDescriptionOfAWonTrophy() {
+        val statisticsAndTrophies = composeTestRule.onNodeWithTag("statisticsAndTrophies")
+        statisticsAndTrophies.onChildren().filter(hasTestTag("Trophy${user.trophiesWon[1]}"))[0]
```
Jharaxus, line 40: - user.trophiesWon[1] could be turned into a variable to improve clarity
- using [0] isn't very explicit, if the idea is to test only one trophy then it should be explained through a comment to make it clearer

app/src/androidTest/java/com/github/polypoly/app/menu/ProfileActivityTest.kt
```diff
@@ -0,0 +1,96 @@
...
+    @Test
+    fun canSeeTheCorrectNumberOfTrophiesWon() {
+        composeTestRule.onNodeWithText(user.trophiesWon.size.toString()).performScrollTo().assertIsDisplayed()
+    }
+
+    @Test
+    fun canSeeTheChosenDisplayedTrophies() {
+        user.trophiesDisplay.forEach {
+            val profileSurface = composeTestRule.onNodeWithTag("profileSurface")
+            profileSurface.onChildren().filter(hasTestTag("Trophy${it}")).assertCountEquals(1)
```
Jharaxus, line 78: ${it} doesn't require the brackets { and } when the element interpolated is one-word

app/src/main/java/com/github/polypoly/app/menu/ProfileModifyingActivity.kt
```diff
@@ -104,45 +90,144 @@ class ProfileModifyingActivity : ComponentActivity() {
...
+            textAlign = TextAlign.Center)
+        Spacer(modifier = Modifier.height(20.dp))
+        TrophiesView(
+            callBack = { idx ->
+                if(!trophiesDisplay.contains(idx) && user.hasTrophy(idx)) {
+                    if (trophiesDisplay.size >= 3) trophiesDisplay.removeAt(0)
+                    trophiesDisplay.add(idx)
+                }
+            },
+            maxSelected = 3,
```
Jharaxus, line 110: 3 here could be turned into a variable to keep consistency with the `trophiesDisplay.size >= 3` above

app/src/main/java/com/github/polypoly/app/menu/shared_component/TrophiesView.kt
```diff
@@ -0,0 +1,132 @@
...
+            .advancedShadow(
+                alpha = if (selected && !disable) 0.3f else 0f,
+                color = MaterialTheme.colors.onSecondary,
+                shadowBlurRadius = PADDING_TROPHIES.dp,
+                cornersRadius = TROPHY_SIZE.dp)
+            .clip(CircleShape)
+            .background(
+                color = trophyColor)
+            .background(
+                color = Color.White.copy(alpha = if (selected) 0f else 0.6f))
```
Jharaxus, line 61: `Color.White` could be replaced with `MaterialTheme.colors.background` to improve its meaningfulness

Jharaxus (changes requested):
A few little things to change.
I didn't mention the magic numbers used in the UI as, similarly to [#55](https://github.com/polypoly-team/android-app/pull/55), they should be dealt with in [this issue](https://github.com/polypoly-team/android-app/issues/40).

app/src/androidTest/java/com/github/polypoly/app/menu/ProfileActivityTest.kt
tamaraaPrs: after the filter, there are only one trophy but in a list, so I put [0] to unwraped it

app/src/androidTest/java/com/github/polypoly/app/menu/ProfileModifyingActicityTest.kt
```diff
@@ -96,4 +96,24 @@ class ProfileModifyingActivityTest {
 
         Intents.release()
     }
+
+    @Test
+    fun cantValidateTheProfileIfThePlayerGiveAnEmptyNickName() {
```
AlSchlo, line 101: typo

PR #57, opened 2023-03-31
Change Code Climate argument count threshold to 6

jadtl:
This PR changes the argument count threshold of Code Climate from 4 to 6.


Jharaxus (approved):
This should work as expected

PR #59, opened 2023-03-31
Add link to MapView

AmLibra:
Self explanatory


app/src/main/java/com/github/polypoly/app/menu/WelcomeActivity.kt
```diff
@@ -138,13 +140,21 @@ class WelcomeActivity : ComponentActivity() {
         ) {
             ProfileOptionButton()
             RulesOptionButton()
-            OptionButton(
-                onClick = { /*TODO*/ },
-                icon_id = R.drawable.tmp_happysmile,
-                description = "optionButton2")
+            MapButton()
```
jadtl, line 143: This should only be a temporary button because it doesn't make sense to have access to the map without any game set up. Maybe this should be annotated?

jadtl (changes requested):
Only a small thing, otherwise looks good!

jadtl (approved):

PR #73, opened 2023-04-14
Update README and remove .idea

AmLibra:
Functionality stays the same, no features added.


gradle.properties
```diff
@@ -20,4 +20,6 @@ kotlin.code.style=official
 # Enables namespacing of each library's R class so that its R class includes only the
 # resources declared in the library itself and none from the library's dependencies,
 # thereby reducing the size of the R class for that library
-android.nonTransitiveRClass=true
\ No newline at end of file
+android.nonTransitiveRClass=true
+android.defaults.buildfeatures.buildconfig=true
+android.nonFinalResIds=false
```
Jharaxus, line 25: I am not convinced that these changes should be in this pr, but they look benign so it should be ok

Jharaxus (approved):
No big changes here, this looks fine for a merge :)

PR #74, opened 2023-04-14
Add most locations on EPFL and detect when near one of them

AmLibra:
Self-explicit:
- most important EPFL buildings added
- refactoring of the map implementation
- a text pops up informing the player when they are near enough of a building to interact with it.


.codeclimate.yml
```diff
@@ -2,9 +2,21 @@ exclude_patterns:
...
+    method-lines:
+      config:
+        threshold: 35
+    similar-code:
+      enabled: false
+    identical-code:
+      enabled: false
+    file-lines:
+      config:
+        threshold: 500
```
Jharaxus, line 22: This should be done in a separate pr next time as it is not part of this pr's subject

app/src/androidTest/java/com/github/polypoly/app/map/LocationRepositoryTest.kt
```diff
@@ -3,15 +3,15 @@ package com.github.polypoly.app.map
 import org.junit.Test
 
 // These tests are not really useful, but they're here symbolically, this is only a data class
-class LocalizationRepositoryTest {
+class LocationRepositoryTest {
     @Test
     fun testGetLocalizationsWorks() {
-        assert(LocalizationRepository.getZones().flatMap { zone -> zone.localizations }
+        assert(LocationRepository.getZones().flatMap { zone -> zone.locations }
             .isNotEmpty())
```
Jharaxus, line 10: This can be refactored into `assert(LocationRepository.getZones().flatMap(Zone::locations).isNotEmpty())` (usually `{x -> x.xyz}` can be replaced by `(X::xyz)` to make it more concise and more expressive, where `xyz` is a an attribute or a member function of the class `X`).

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -69,7 +87,9 @@ class MapActivity : ComponentActivity() {
     private val initialZoom = 18.0
     private val markerSideLength = 100
 
-    private val markerToLocalization = mutableMapOf<Marker, Localization>()
+    private val maxCloseLocationDistance = 10.0
```
Jharaxus, line 90: Private constants such as this one (or above) would perhaps be more expressive if named using capital snake-case and put into the class' companion object (kinda like static constants in other languages).

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -88,7 +108,11 @@ class MapActivity : ComponentActivity() {
                 ) {
                     MapView()
                     BuildingInfoUIComponent()
-                    Hud(PlayerGlobalData(false, 420), listOf(PlayerGlobalData(false, 32), PlayerGlobalData(false, 56)),16)
+                    Hud(
+                        PlayerGlobalData(false, 420),
+                        listOf(PlayerGlobalData(false, 32), PlayerGlobalData(false, 56)),
+                        16
+                    )
```
Jharaxus, line 115: I belief these PlayerGlobalData are placeholders until therer is a way to retrieve the real ones. If so, they should be declared separately and explicitly named as such instead of being inlined within the code as it is right now (it leaves magic constants coming from nowhere and is not very meaningful)

.codeclimate.yml
AmLibra: While these modifications do not directly address the subject of the PR, they are still related as the changes in the PR conflict with the "current" ruleset of codeclimate.
"current" is in quotation marks, as these are changes that I have already made in my previous PR following the group's approval. I can find them as early as commit f41ec00d0c71f235a731bf70e4ed5ad987624d49, from 2 weeks ago!

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -105,9 +129,9 @@ class MapActivity : ComponentActivity() {
...
 
             for (zone in getZones())
-                zone.localizations
+                zone.locations
                     .forEach {
-                        markerToLocalization[addMarkerTo(
+                        markerToLocation[addMarkerTo(
                             mapView,
                             it.position,
                             it.name,
```
Jharaxus, line 137: Why use a foreach here? You could simply use an inner for-loop `for (location in zone.locations) {...}`, which is more conventional and less verbose than a foreach.
Moreover, when using lambdas in kotlin, the implicit lambda argument `it` is not very expressive (even if it is convenient as it is implicitly defined), so it's usually better to rename it appropriately.
Finally, the Marker creation and its assignation in the mapping should be done in two separate lines as it is a bit heavy right now.

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -360,42 +415,74 @@ class MapActivity : ComponentActivity() {
...
         return mapView.overlays.filterIsInstance<Marker>()
     }
 
-    private fun updateAllDistances(mapView: MapView, myLocation: GeoPoint) {
+    /**
+     * Updates the distance of all markers and returns the closest one.
+     *
+     * @return the closest location or null if there are no locations close enough to the player
+     */
+    private fun updateAllDistancesAndClosest(
```
Jharaxus, line 462: Maybe this could be renamed into `updateAllDistancesAndFindClosest` to make it clearer.
However, if there is a need to add an _and_ in the function name, that usually means that function does too many things. Here, is there a sufficient performance gain to do the closest lookup while doing the update, compared to doing this in two separate functions?

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -360,42 +415,74 @@ class MapActivity : ComponentActivity() {
...
         val locationOverlay = object : MyLocationNewOverlay(locationProvider, mapView) {
             override fun onLocationChanged(location: Location?, provider: IMyLocationProvider?) {
                 super.onLocationChanged(location, provider)
-                updateAllDistances(mapView, GeoPoint(location))
+                mapViewModel.setCloseLocation(
+                    updateAllDistancesAndClosest(mapView, GeoPoint(location))
+                )
                 mapViewModel.addDistanceWalked(lastLocation.distanceTo(location!!))
                 lastLocation = locationProvider.lastKnownLocation
+                print("location: $location")
```
Jharaxus, line 433: Is it a forgotten debug print?

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -360,42 +415,74 @@ class MapActivity : ComponentActivity() {
...
             }
         }
 
         locationOverlay.enableMyLocation()
         locationOverlay.enableFollowLocation()
         locationOverlay.runOnFirstFix {
+            print("First fix: ${locationOverlay.myLocation}")
```
Jharaxus, line 440: Is it another forgotten debug print?

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
Jharaxus, line 90: I do not quite get what that variable is used for just with its name --> could you maybe come up with a better one?

Jharaxus (changes requested):

app/src/androidTest/java/com/github/polypoly/app/map/LocationRepositoryTest.kt
AmLibra: I sure wish I could as a Scala enjoyer, but I found a neater workaround, hope it's good enough:
assert(LocationRepository.getZones().flatMap { it.locations }.isNotEmpty())

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
AmLibra: Not my PR!

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
AmLibra: Addressed!

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
AmLibra: Addressed

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
AmLibra: I guess that 2n is still n so technically no, but in my brain that is a lot for a phone. If you really insist then I can split it into 2 different functions. No real opinion as both cases (long name and bad complexity) feel just as bad to me.

AmLibra: Name changed as yours is definitely better

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
AmLibra: I write so little debug prints in this project due to the nature of the bugs that i forgot that one, thanks!

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
AmLibra: ditto above

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
AmLibra: a comment should do the trick, or else the name would be way too long even for a static constant

.codeclimate.yml
Jharaxus: Ok that makes sense! Actually it is simply an advise for next time, nothing crucial here.

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
Jharaxus: Well the issue is that line 433 you have `updateAllDistancesAndFindClosest(mapView, locationOverlay.myLocation)` which doesn't use the closest location found, which to me reveals that you have two different behaviors here that you could separate into `updateDistances` and `findClosestLocation`.
On the performance side it just takes profit of a for-loop only ran once instead of twice so it should be benign considering we don't have many locations.

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -126,18 +121,13 @@ class MapActivity : ComponentActivity() {
         AndroidView(factory = { context ->
             Configuration.getInstance().userAgentValue = BuildConfig.APPLICATION_ID
 
-            val mapView = initMapView(context, initialPosition)
+            val mapView = initMapView(context, Companion.INITIAL_POSITION)
```
Jharaxus, line 124: You can remove the `Companion` reference (it is inferred)

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -369,7 +359,7 @@ class MapActivity : ComponentActivity() {
         mapView.setTileSource(TileSourceFactory.DEFAULT_TILE_SOURCE)
         mapView.setMultiTouchControls(true)
         mapView.zoomController.setVisibility(CustomZoomButtonsController.Visibility.NEVER)
-        mapView.controller.setZoom(initialZoom)
+        mapView.controller.setZoom(Companion.INITIAL_ZOOM)
```
Jharaxus, line 362: Same `Companion` reference is useless

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -785,4 +777,13 @@ class MapActivity : ComponentActivity() {
...
         }
     }
+
+    companion object {
+        private val INITIAL_POSITION = GeoPoint(46.518726, 6.566613)
+        private const val INITIAL_ZOOM = 18.0
+        private const val MARKER_SIDE_LENGTH = 100
+        private const val MAX_CLOSE_LOCATION_DISTANCE = 10.0    // meters, used to determine
+                                                                // if the player is close enough to
+                                                                //a location to interact with it
```
Jharaxus, line 787: An entire one line comment would be better than this 3 lines one that tries to match the 80 max char per line

Jharaxus (changes requested):
A couple of useless `Companion` references can be removed (I only marked the first two to avoid overweighing the review).

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
Jharaxus: Ok you simply fixed the indentation so this can be left to a later refactoring if needed as it is nothing big.

app/src/androidTest/java/com/github/polypoly/app/map/LocationRepositoryTest.kt
Jharaxus: Fellow Scala enjoyer, this should be doable in kotlin too, the code I sent above `assert(LocationRepository.getZones().flatMap(Zone::locations).isNotEmpty()) (usually {x -> x.xyz}` works on my machine 😄

AmLibra: Indeed seems to work, IDE probably needed a good reboot

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
AmLibra: The mistake was that runOnFirstFix was not using the output, accidental omission on my end.

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
AmLibra: As you so wish

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
Jharaxus: Ok this makes more sense now 👍

app/src/androidTest/java/com/github/polypoly/app/map/LocationRepositoryTest.kt
Jharaxus: Great 👍

Jharaxus (approved):
This works for the merge :)

app/src/androidTest/java/com/github/polypoly/app/map/LocationRepositoryTest.kt
```diff
@@ -1,17 +1,17 @@
...
 import org.junit.Test
 
 // These tests are not really useful, but they're here symbolically, this is only a data class
-class LocalizationRepositoryTest {
+class LocationRepositoryTest {
     @Test
     fun testGetLocalizationsWorks() {
-        assert(LocalizationRepository.getZones().flatMap { zone -> zone.localizations }
-            .isNotEmpty())
+        assert(LocationRepository.getZones().flatMap(Zone::locations).isNotEmpty())
```
SolalPirelli, line 10: excellent example of cleaning up code while you're at it :)

app/src/main/java/com/github/polypoly/app/map/LocationRepository.kt
```diff
@@ -0,0 +1,459 @@
...
+        // Pastel lime
+        0xFFDCE775.toInt(),
+        // Pastel teal
+        0xFF4DB6AC.toInt(),
+        // Pastel indigo
+        0xFF7986CB.toInt()
+    )
+
+    /**
+     * List of localizations in the EPFL campus area.
```
SolalPirelli, line 44: one missed localization->location here

app/src/main/java/com/github/polypoly/app/map/LocationRepository.kt
```diff
@@ -0,0 +1,459 @@
...
+     * List of zones in the EPFL campus area. Each zone has a color associated with it.
+     */
+    private val zones = listOf(
+        Zone( // BC, INN, INR
+            locations = listOf(
+                locations[0],
+                locations[4],
+                locations[5]
+            ),
+            color = colours[0]
```
SolalPirelli, line 353: mix of US and UK spelling

app/src/main/java/com/github/polypoly/app/map/LocationRepository.kt
```diff
@@ -0,0 +1,459 @@
...
+                locations[39],
+                locations[40],
+                locations[41]
+            ),
+            color = colours[12]
+        )
+    )
+
+    /**
+     * Returns the list of localizations.
```
SolalPirelli, line 456: another missed replace

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -65,11 +83,8 @@ class MapActivity : ComponentActivity() {
 
     // Not public for testing purposes
     val mapViewModel: MapViewModel = MapViewModel()
-    private val initialPosition = GeoPoint(46.518726, 6.566613)
-    private val initialZoom = 18.0
-    private val markerSideLength = 100
 
-    private val markerToLocalization = mutableMapOf<Marker, Localization>()
+    private val markerToLocation = mutableMapOf<Marker, com.github.polypoly.app.game.Location>()
```
SolalPirelli, line 87: why use the full name?

app/src/androidTest/java/com/github/polypoly/app/map/LocationRepositoryTest.kt
AlSchlo: It's still the wrong assert function though

PR #77, opened 2023-04-15
Feat/authentication

Kaz-ookid:
adds specific **WelcomeActivity** UI when user is not signed in: there only is a "**sign in**" button displayed, and the logo picture.

The button launches **google authentication services**. 

Once the user is logged in, the normal **WelcomActivity** menu is displayed. User can sign out in the settings with a dedicated button, and is sent back to the **WelcomeActivity**.

Still figuring out how to test google authentication.


app/src/main/java/com/github/polypoly/app/WelcomeActivity.kt
```diff
@@ -2,36 +2,72 @@ package com.github.polypoly.app
...
+import com.firebase.ui.auth.AuthUI
+import com.firebase.ui.auth.FirebaseAuthUIActivityResultContract
+import com.firebase.ui.auth.data.model.FirebaseAuthUIAuthenticationResult
 import com.github.polypoly.app.menu.JoinGameLobbyActivity
 import com.github.polypoly.app.menu.MenuComposable
 import com.github.polypoly.app.menu.kotlin.GameMusic
-
 import com.github.polypoly.app.ui.theme.PolypolyTheme
+import com.google.firebase.auth.FirebaseAuth
+import androidx.activity.ComponentActivity as ComponentActivity1
```
vrooomer, line 30: Why did you import this as `ComponentActivity1`?

app/src/main/java/com/github/polypoly/app/menu/SettingsActivity.kt
```diff
@@ -31,6 +33,7 @@ class SettingsActivity : MenuActivity("Settings") {
         setContent {
             MenuContent {
                 SettingsContent()
+                SignOutButton()
```
vrooomer, line 36: You should put `SignOutButton()` in the code of `SettingsContent`, `MenuContent` is made to contain only a global @Composable, even if this works, it's better to stick to the mentionned convention :)

app/src/main/java/com/github/polypoly/app/WelcomeActivity.kt
```diff
@@ -43,6 +79,18 @@ class WelcomeActivity : ComponentActivity() {
     fun WelcomeContent() {
         GameMusic.setSong(LocalContext.current, R.raw.mocksong)
         GameMusic.startSong()
+
+        val isSignedIn = remember { mutableStateOf(false) }
+        mAuthListener = FirebaseAuth.AuthStateListener {
+            val user = FirebaseAuth.getInstance().currentUser
+            isSignedIn.value = user != null
+            if(isSignedIn.value) {
+                print("\n__________________ Signed in: ${user?.displayName} _________________\n")
```
vrooomer, line 88: Are these prints used for debug? Maybe you should remove them (when the PR is done)

vrooomer (changes requested):
The code looks great overall! As this PR is still a draft, i'll request changes to let you finish it and check the few comments that I made.

~ v:)

app/src/main/java/com/github/polypoly/app/WelcomeActivity.kt
Kaz-ookid: just an error, thanks for pointing it out!

app/src/main/java/com/github/polypoly/app/menu/SettingsActivity.kt
Kaz-ookid: done, thanks for the tip :)!

app/src/main/java/com/github/polypoly/app/WelcomeActivity.kt
Kaz-ookid: yes they are, I will remove them at the very end

app/src/main/java/com/github/polypoly/app/game/GameLobby.kt
```diff
@@ -1,30 +1,28 @@
...
+    val gameMap: List<Zone> = listOf(),
+    val initialPlayerBalance: Int = 0,
+    val name: String = "default-lobby-instance",
+    val code: String = "default-lobby-code",
     val private: Boolean = false
 ) {
 
-    private val currentUsersRegistered: ArrayList<User> = ArrayList()
-    val usersRegistered: List<User> get() = currentUsersRegistered.toList()
+    val usersRegistered: ArrayList<User> = ArrayList()
```
vrooomer, line 18: are you sure that removing this getter won't break immutability?

app/src/main/java/com/github/polypoly/app/menu/JoinGameLobbyActivity.kt
```diff
@@ -32,28 +32,35 @@ import com.github.polypoly.app.game.GameLobby
...
 import timber.log.Timber
+import java.util.concurrent.CompletableFuture
 
 /**
  * Activity where the user can join a gameLobby
  */
-class JoinGameLobbyActivity : ComponentActivity() {
+class JoinGameLobbyActivity : MenuActivity("Join a game") {
+    companion object {
+        const val POLLING_INTERVAL = 5000L
```
vrooomer, line 48: Are you sure you want to do polling instead of using a listener?

app/src/main/java/com/github/polypoly/app/menu/JoinGameLobbyActivity.kt
```diff
@@ -130,6 +137,9 @@ class JoinGameLobbyActivity : ComponentActivity() {
         var text by remember { mutableStateOf("") }
 
         OutlinedTextField(
+            colors = TextFieldDefaults.outlinedTextFieldColors(
```
vrooomer, line 140: You should use the defined text field colors `UIElements.outlineTextFieldColors()`

app/src/main/java/com/github/polypoly/app/menu/JoinGameLobbyActivity.kt
```diff
@@ -32,28 +32,35 @@ import com.github.polypoly.app.game.GameLobby
...
 
+    private val fakeAuthenticatedUser = User(7, "fake_user", "I am fake until google authentication is setup", Skin(0, 0, 0),
+        Stats(0, 0, 0, 0, 0), listOf(), mutableListOf()
+    )
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContent {
-            PolypolyTheme {
+            MenuContent{
```
vrooomer, line 63: Once again, let's stick to the convention of having a composable that contains all the UI named `ActivityNameContent` so that the onCreate function isn't that big.

app/src/main/java/com/github/polypoly/app/menu/JoinGameLobbyActivity.kt
```diff
@@ -482,73 +491,61 @@ class JoinGameLobbyActivity : ComponentActivity() {
...
+    private fun joinGameLobbyRoom() {
         val currentLobbyKey = DB_GAME_LOBIES_PATH + gameLobbyCode
-        val gameLobby =  mockDb.getValue<GameLobby>(currentLobbyKey).get()
-        val newGameLobby = GameLobby(gameLobby.admin, gameLobby.gameMode, gameLobby.minimumNumberOfPlayers, gameLobby.maximumNumberOfPlayers, gameLobby.roundDuration
-            , gameLobby.gameMap, gameLobby.initialPlayerBalance, gameLobby.name, gameLobby.code, gameLobby.private)
-
-        for (player in gameLobby.usersRegistered) {
-            if (player != newGameLobby.admin) {
-                newGameLobby.addUser(player)
+        remoteDB.getValue<GameLobby>(currentLobbyKey).thenAccept { gameLobby ->
```
vrooomer, line 531: You could just do something like
`if(gameLobby.usersRegistered.contains(player) then gameLobby.addUser(player)`

vrooomer (changes requested):
Nice job Kaz!
The overall code looks very nice, however while looking at the DB code, it is clear that we all need to find a common way to work with it because the code related to it is very messy but here it is not the purpose of the PR.

Let me know if some comments are unclear!

~ v:)

Kaz-ookid:
> Nice job Kaz! The overall code looks very nice, however while looking at the DB code, it is clear that we all need to find a common way to work with it because the code related to it is very messy but here it is not the purpose of the PR.
> 
> Let me know if some comments are unclear!
> 
> ~ v:)

I wonder what you are talking about by `while looking at the DB code`, since most of it is coming from #82? Maybe it should be commented there

app/src/main/java/com/github/polypoly/app/menu/JoinGameLobbyActivity.kt
Kaz-ookid: this code came from #82. Actually, he wrote a TODO when polling happens, saying that we should at an end in fact use listeners :)!

app/src/main/java/com/github/polypoly/app/WelcomeActivity.kt
```diff
@@ -17,21 +17,41 @@ import androidx.compose.ui.platform.LocalContext
...
  * These actions may be: creating a game, joining a game, logging in, settings, rules, leaderboards etc.
  */
-class WelcomeActivity : ComponentActivity() {
+class WelcomeActivity : ComponentActivity(){
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContent { WelcomeContent() }
+        if (!isSignedIn) {
+            finish()
```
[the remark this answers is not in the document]
Kaz-ookid: You might be right, I think this might be an **artifact** of when I used to implement this the other way around (**WelcomeActivity** was the main, and **Signin** was called when needed). I will check it in detailed, in case there was a real reason why I left it here.

Also for further explanation on why **SigninActivity** finishes,  **SigninActivity** in fact always finishes when passing to **WelcomeActivity**. However, the only case where **WelcomeActivity** finishes itself is when we sign off the user from **Settings**. What happens in **Settings** is that it signs the user off, launches the **SigninActivity** from intent, then finishes itself, which will cause the **WelcomeActivity** to finish itself too right after that, by readingthe `onResume `function. Then the SigninActivity is launched :)

app/src/main/java/com/github/polypoly/app/SignInActivity.kt
```diff
@@ -0,0 +1,239 @@
...
+                }
+            }
+        }
+    }
+
+    /**
+     * DEBUG function
+     * Add fake data (possibly duplicate from tests' fake data) if the data in DB were corrupted
+     */
+    fun addFakeDataToDB() {
```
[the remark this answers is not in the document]
Kaz-ookid: this comes from #82

app/src/main/java/com/github/polypoly/app/SignInActivity.kt
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    /**
+     * starts the WelcomeActivity and sets the isSignedIn flag to true
+     */
+    private fun launchWelcome(){
+        val welcomeActivityIntent = Intent(this, WelcomeActivity::class.java)
+        isSignedIn = true
+        startActivity(welcomeActivityIntent)
+        finish()
```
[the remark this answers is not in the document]
Kaz-ookid: It looks like it does but I may turn it the other way around just in case

app/src/main/java/com/github/polypoly/app/menu/JoinGameLobbyActivity.kt
Kaz-ookid: This will be part of my undone task that I will carry over next sprint, but thanks for the reminder!

app/src/main/java/com/github/polypoly/app/menu/JoinGameLobbyActivity.kt
Kaz-ookid: this comes from #82 and I thus did not check it yet,  but i'll come back to this on my task for next sprint, thanks :)

app/src/main/java/com/github/polypoly/app/WelcomeActivity.kt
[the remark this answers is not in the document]
vrooomer: Okay I understand better! If it works then it is perfect, thanks for the explanation :)

vrooomer:
About the DB, I didn't know it wasn't your code, oops! Then this code should be fixed in further a PR though.

vrooomer (approved):

app/src/androidTest/java/com/github/polypoly/app/JoinGameLobbyActivityTest.kt
```diff
@@ -20,24 +20,28 @@ class JoinGameLobbyActivityTest: PolyPolyTest(false, true) {
...
+        Thread.sleep(1000)
         composeTestRule.onNodeWithTag("logo").assertIsDisplayed()
     }
 
     @Test
     fun inputInvalidGameLobbyCode_displayWarningMessage() {
         //TODO: Check for a group code that is not in the DB once we have the queries set
-        composeTestRule.onNodeWithTag("gameLobbyCodeField").performTextInput("polypoly")
+        composeTestRule.onNodeWithTag("gameLobbyCodeField").performTextInput("polpolu")
+        Thread.sleep(1000)
```
AlSchlo, line 32: :face_with_spiral_eyes:

app/src/androidTest/java/com/github/polypoly/app/commons/PolyPolyTest.kt
```diff
@@ -28,16 +35,22 @@ import kotlin.time.DurationUnit
...
+    private val fillWithFakeData: Boolean, //> fill remote storage with fake data at the beginning of every test
+    val signFakeUserIn: Boolean = false //> sign a fake user in at the beginning of every test
 ) {
     companion object {
         // Global tests constants
         const val TIMEOUT_DURATION = 5L
 
+        // Ensures only one global initialization even for multithreaded testing
+        var globalInitCompleted = false
+        var initLock = ReentrantLock(true)
```
AlSchlo, line 47: Where is this variable even used?

app/src/main/java/com/github/polypoly/app/SignInActivity.kt
```diff
@@ -0,0 +1,241 @@
...
+    /**
+     * launched when the sign-in flow is wanted to be started
+     */
+    private val signInLauncher = registerForActivityResult(
+        FirebaseAuthUIActivityResultContract()
+    ) {}
+
+    @Composable
+    fun SignInContent() {
+        GameMusic.setSong(LocalContext.current, R.raw.mocksong)
```
AlSchlo, line 94: Why is the music launched in the sign in activity? This seems a weird Software Engineering practice

app/src/androidTest/java/com/github/polypoly/app/JoinGameLobbyActivityTest.kt
Kaz-ookid: This was some hopeless try to make the test pass in cirrus, since I had no clue about Solal’s solution yet :( we all had this problem of tests not running and the common solution was to comment the tests with issues, which I did not like as a solution. I prefere to have this dirty solution for now, while the tests are still getting checked. Also it is marked as an issue to resolve these tests and I believe we talked about turning it into a task later on

app/src/androidTest/java/com/github/polypoly/app/commons/PolyPolyTest.kt
Kaz-ookid: This comes from the main while merging to resolve conflicts, it is not my code

app/src/main/java/com/github/polypoly/app/SignInActivity.kt
Kaz-ookid: You are right it was not intended. I think I have put it there because Signing activity is the main activity now, but it makes sense for the music to launch only when we reach the menu.

PR #82, opened 2023-04-18
Feature/connect with real db

Jharaxus:
Remove `FakeRemoveStorage` and connect the App with the real database instead.


app/src/androidTest/java/com/github/polypoly/app/commons/PolyPolyTest.kt
```diff
@@ -0,0 +1,144 @@
...
+import com.github.polypoly.app.network.RemoteDB
+import com.google.firebase.database.DatabaseReference
+import com.google.firebase.database.ktx.database
+import com.google.firebase.ktx.Firebase
+import org.junit.Before
+import org.junit.runner.RunWith
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.locks.ReentrantLock
+import kotlin.time.Duration.Companion.hours
```
Kaz-ookid, line 19: unused imports

app/src/main/java/com/github/polypoly/app/menu/JoinGameLobbyActivity.kt
Kaz-ookid, line 1: Nice update to the activity! However some functionalities have broken after your changes. I am just writing it here for the record, but I will repare all this in my next PR, as I am supposed to work on this class :)!

That is why i will not further comment this file.

app/src/main/java/com/github/polypoly/app/WelcomeActivity.kt
```diff
@@ -31,9 +46,84 @@ import com.github.polypoly.app.ui.theme.PolypolyTheme
...
+        addFakeDataToDB() //> Uncomment this line if you need to refresh the fake data within the DB
+
         setContent { WelcomeContent() }
     }
 
+    /**
+     * DEBUG function
+     * Add fake data (possibly duplicate from tests' fake data) if the data in DB were corrupted
+     */
+    fun addFakeDataToDB() {
```
vrooomer, line 62: To we need to keep a debug function into our final code?

app/src/main/java/com/github/polypoly/app/game/GameLobby.kt
```diff
@@ -1,30 +1,28 @@
...
-    val maximumNumberOfPlayers: Int,
-    val roundDuration: Duration,
-    val gameMap: List<Zone>,
-    val initialPlayerBalance: Int,
-    val name: String,
-    val code: String,
+data class GameLobby(
+    val admin: User = User(),
+    val gameMode: GameMode = GameMode.LAST_STANDING,
+    val minimumNumberOfPlayers: Int = 2,
```
vrooomer, line 8: Be careful when merging this because in the new version we have a new class that contains all the rules, `GameRules`

app/src/main/java/com/github/polypoly/app/game/GameLobby.kt
```diff
@@ -1,30 +1,28 @@
...
+    val gameMap: List<Zone> = listOf(),
+    val initialPlayerBalance: Int = 0,
+    val name: String = "default-lobby-instance",
+    val code: String = "default-lobby-code",
     val private: Boolean = false
 ) {
 
-    private val currentUsersRegistered: ArrayList<User> = ArrayList()
-    val usersRegistered: List<User> get() = currentUsersRegistered.toList()
+    val usersRegistered: ArrayList<User> = ArrayList()
```
vrooomer, line 18: Are you sure that changing this won't break the immutability?

app/src/main/java/com/github/polypoly/app/menu/JoinGameLobbyActivity.kt
```diff
@@ -32,25 +33,33 @@ import com.github.polypoly.app.game.GameLobby
...
 import kotlinx.coroutines.delay
 import timber.log.Timber
+import java.util.concurrent.CompletableFuture
 
 /**
  * Activity where the user can join a gameLobby
  */
 class JoinGameLobbyActivity : MenuActivity("Join Game") {
+    companion object {
+        const val POLLING_INTERVAL = 5000L
```
vrooomer, line 51: Maxime said in #77 that this will be changed, because polling is worse than listening, do you agree?

app/src/main/java/com/github/polypoly/app/menu/JoinGameLobbyActivity.kt
```diff
@@ -491,73 +499,61 @@ class JoinGameLobbyActivity : MenuActivity("Join Game") {
...
+    private fun joinGameLobbyRoom() {
         val currentLobbyKey = DB_GAME_LOBIES_PATH + gameLobbyCode
-        val gameLobby =  mockDb.getValue<GameLobby>(currentLobbyKey).get()
-        val newGameLobby = GameLobby(gameLobby.admin, gameLobby.gameMode, gameLobby.minimumNumberOfPlayers, gameLobby.maximumNumberOfPlayers, gameLobby.roundDuration
-            , gameLobby.gameMap, gameLobby.initialPlayerBalance, gameLobby.name, gameLobby.code, gameLobby.private)
-
-        for (player in gameLobby.usersRegistered) {
-            if (player != newGameLobby.admin) {
-                newGameLobby.addUser(player)
+        remoteDB.getValue<GameLobby>(currentLobbyKey).thenAccept { gameLobby ->
```
vrooomer, line 539: As I said in #77, isn't it a bit too much? I mean you could do something like:
`if(!lobby.users.contains(currentUser)) then lobby.addUser(currentUser)`

vrooomer (changes requested):
Nice job on this one Jharaxus!

It's nice to finally be able to use the real DB, the implementation seems easy to use and straight forward. However I pointed small issues, you should look at them even if they're small!

~ v:)

app/src/main/java/com/github/polypoly/app/WelcomeActivity.kt
Jharaxus: Now that we are connected to the real database, we need it to be populated with real values so that the application can make sense (especially now that not all feature exist yet). For example, when going to the JoinGameLobby activity,  if no game lobby exist yet in the DB then nothing happen in the activity (all codes fail and no gamelobby is shown in the list of existing ones). The same happens when going to the Profile activity, and this is valid for handmade testing, ongoing development and the live demos.
This is the reason why I added this "debug" function (maybe not the best adjective) to populate the DB at launch to ensure we have interesting data to work with (it may be corrupted between two App launches).
This has the downside of preventing persistency, but unit tests already test that it works and the current app state makes that acceptable.

app/src/main/java/com/github/polypoly/app/game/GameLobby.kt
Jharaxus: Yes this will be fixed when merging the other branch on main

app/src/main/java/com/github/polypoly/app/WelcomeActivity.kt
vrooomer: But couldn't we just populate the DB once and then just get elements there?

app/src/main/java/com/github/polypoly/app/game/GameLobby.kt
Jharaxus: There is indeed a tradeoff that I made here: the initial version had better encapsulation as no outside class could directly add/remove users (they had to go through the specific functions), however I had to turn it that way because firebase expects to be able to directly set the users with an existing list of Users when fetching a User from the DB. With the previous version it didn't work :(

app/src/main/java/com/github/polypoly/app/menu/JoinGameLobbyActivity.kt
Jharaxus: Yes I also added a TODO comment line 213 (where it is used) as it is only the intermediate implementation until event listening is implemented with firebase.

app/src/main/java/com/github/polypoly/app/game/GameLobby.kt
vrooomer: Oh okay! Then it's a faire tradeoff :)

app/src/main/java/com/github/polypoly/app/menu/JoinGameLobbyActivity.kt
Jharaxus: Nice one! I didn't even check that the user is not contained as otherwise the user would already be in the game.

app/src/main/java/com/github/polypoly/app/menu/JoinGameLobbyActivity.kt
```diff
@@ -537,20 +537,8 @@ class JoinGameLobbyActivity : MenuActivity("Join Game") {
...
-            for (player in gameLobby.usersRegistered) {
-                if (player != newGameLobby.admin) {
-                    newGameLobby.addUser(player)
-                }
-            }
-
-            newGameLobby.addUser(fakeAuthenticatedUser)
-            remoteDB.updateValue(currentLobbyKey, newGameLobby)
+            gameLobby.addUser(fakeAuthenticatedUser)
+            remoteDB.updateValue(currentLobbyKey, gameLobby)
```
vrooomer, line 541: Perfect! You don't even need to update the DB as this will be done in `GameLobby` ;)

app/src/main/java/com/github/polypoly/app/WelcomeActivity.kt
Jharaxus: No because they can easily be corrupted through wrong manipulation (I personally experienced it a few time in this sprint 😢) and prevent repeatability when developing new features. Moreover as of today we all use the same DB, so data can cross over if 2 people successively their code which may lead to issues. For example is a class gets new fields in the code of 1 person, if an other person run his code later than the data will be corrupted on his point of view: refreshing the data every time solves that.
I do not believe this should necessarily stay for a lot of sprints, but at least for the next one this should come be useful.

vrooomer (approved):

app/src/androidTest/java/com/github/polypoly/app/commons/PolyPolyTest.kt
```diff
@@ -0,0 +1,157 @@
...
+import kotlin.time.DurationUnit
+
+@RunWith(AndroidJUnit4::class)
+abstract class PolyPolyTest(
+    private val clearRemoteStorage: Boolean, //> clear remote storage at the beginning of every test
+    private val fillWithFakeData: Boolean //> fill remote storage with fake data at the beginning of every test
+) {
```
SolalPirelli, line 29: This is a bit confusing, first because they're Booleans so just reading the code it's not clear what, e.g., `PolyPolyTest(true, false)` means, but also why would a test want the `false, true` option? `JoinGameLobbyActivityTest` does this above but what does this mean for the test, since it presumably won't know what other data there is aside from the test data? An enum with three options "leave database unchanged", "clear database", and "set database to fake data" would be more readable and make more sense IMHO.

app/src/main/java/com/github/polypoly/app/WelcomeActivity.kt
```diff
@@ -31,9 +46,84 @@ import com.github.polypoly.app.ui.theme.PolypolyTheme
 class WelcomeActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+
+        val db = Firebase.database
+        remoteDB = RemoteDB(db, "live")
+
+        addFakeDataToDB() //> Uncomment this line if you need to refresh the fake data within the DB
```
SolalPirelli, line 53: given the comment this presumably should've been commented out when merged?

app/src/main/java/com/github/polypoly/app/menu/JoinGameLobbyActivity.kt
```diff
@@ -491,73 +499,49 @@ class JoinGameLobbyActivity : MenuActivity("Join Game") {
...
-        } else if(gameLobbyIsFull(gameLobbyCode)){
-            warningState.value = getString(R.string.game_lobby_is_full)
         } else {
-            warningState.value = ""
-            joinGameLobbyRoom(mContext)
+            val lobbyKey = DB_GAME_LOBIES_PATH + gameLobbyCode
+            remoteDB.keyExists(lobbyKey).thenCompose { keyExists ->
+                if (!keyExists) {
+                    warningState.value = getString(R.string.game_lobby_does_not_exist)
+                    CompletableFuture.failedFuture(IllegalAccessException())
```
SolalPirelli, line 518: this is not a great choice of exception, `IllegalAccessException` is intended for reflection operations, perhaps `NoSuchElementException` instead?

app/src/main/java/com/github/polypoly/app/menu/JoinGameLobbyActivity.kt
```diff
@@ -491,73 +499,49 @@ class JoinGameLobbyActivity : MenuActivity("Join Game") {
...
-            warningState.value = getString(R.string.game_lobby_does_not_exist)
-        } else if(gameLobbyIsFull(gameLobbyCode)){
-            warningState.value = getString(R.string.game_lobby_is_full)
         } else {
-            warningState.value = ""
-            joinGameLobbyRoom(mContext)
+            val lobbyKey = DB_GAME_LOBIES_PATH + gameLobbyCode
+            remoteDB.keyExists(lobbyKey).thenCompose { keyExists ->
+                if (!keyExists) {
+                    warningState.value = getString(R.string.game_lobby_does_not_exist)
```
SolalPirelli, line 517: it's odd that the logic to set `warningState` is split in two, once here and once in the `thenAccept` below

app/src/main/java/com/github/polypoly/app/menu/JoinGameLobbyActivity.kt
```diff
@@ -491,73 +499,49 @@ class JoinGameLobbyActivity : MenuActivity("Join Game") {
...
+                    CompletableFuture.failedFuture(IllegalAccessException())
+                } else {
+                    remoteDB.getValue<GameLobby>(lobbyKey)
+                }
+            }.thenAccept { gameLobby ->
+                if(gameLobbyIsFull(gameLobby)){
```
SolalPirelli, line 523: seems like this file wasn't auto-formatted (you should be able to configure your IDE to do this automatically when saving)

Issue #84, opened 2023-04-18
Technical debt in MapActivity.kt


AmLibra:
As mentioned in this week's PR, the MapActivity class is getting way too large, and I propose the following changes:
- Move more of the backend stuff and data storing to the MapViewModel Class, such as the list of visible "Locations" (see issue 83) and the logic to roll dice.
- Add a class to contain all the low level UI Composable components, leaving only the top-level descriptions in MapActivity, as they are required for the UX and game logic. Another solution for Composable functions would be to create different files for each UI element, e.g. MapView.kt, BuildingInfoUIComponent.kt, and Hud.kt.

SolalPirelli:
Can you frame this in a way that helps users as well? Are there features that are currently too hard to build / too buggy / ... as a result of this debt?

Issue #83, opened 2023-04-18
Location(or Building, or Property) naming


AmLibra:
As mentioned in my pull request this week, the name Location might not be well suited for the MapActivity class, which requires us to juggle with the Location from OSM used to locate the phone user, and Location that is used as a part of the game.

The name that would better suit Monopoly would be Property. This would allow us to differentiate the two classes without adding the import prefix each time, and to remain more faithful to the original game.

jadtl:
I'm concerned that "property" would collide with the term "property" as in "attribute", even though it hasn't been used in our codebase yet. Maybe "building" is a safer choice? It's also more descriptive in our case.

PR #85, opened 2023-04-18
Add Dice Roll Feature

AmLibra:
Could be much more, but I think that we should fix the technical debt before adding more stuff on top of an already bloated file. 

The missing features would include making sure that all 3 dice rolls yield 3 different locations, as well as detecting when the player reaches the target location.
Another issue is that Location class name is not well suited for this class, as it conflicts with the built-in Location type that is used for the Map. Property would be a better suited name.
The statements above will naturally be turned into issues on Github. I will also add an issue for the missing features, but that will be very easy to fix once the refactoring will have happened.
 #83 #84

### DIFF
--- a/app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
+++ b/app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
@@ -31,10 +31,13 @@ import androidx.compose.foundation.shape.CircleShape
 import androidx.compose.foundation.text.KeyboardOptions
 import androidx.compose.material.AlertDialog
 import androidx.compose.material.Button
+import androidx.compose.material.Icon
 import androidx.compose.material.MaterialTheme
 import androidx.compose.material.Surface
 import androidx.compose.material.Text
 import androidx.compose.material.TextField
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Casino
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.getValue
@@ -86,9 +89,12 @@ class MapActivity : ComponentActivity() {
 
     private val markerToLocation = mutableMapOf<Marker, com.github.polypoly.app.game.Location>()
 
-    // flag to show the dialog
+    // flag to show the building info dialog
     val showDialog = mutableStateOf(false)
     lateinit var currentMarker: Marker
+
+    // flag to show the roll dice dialog
+    val showRollDiceDialog = mutableStateOf(false)
 
     // store the map view for testing purposes
     lateinit var mapView: MapView private set
@@ -103,11 +109,13 @@ class MapActivity : ComponentActivity() {
                 ) {
                     MapView()
                     BuildingInfoUIComponent()
+                    RollDiceDialog()
                     Hud(
                         PlayerGlobalData(false, 420),
                         listOf(PlayerGlobalData(false, 32), PlayerGlobalData(false, 56)),
                         16
                     )
+                    RollDiceButton()
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -136,6 +144,73 @@ class MapActivity : ComponentActivity() {
...
+    fun RollDiceDialog() {
+        if (showRollDiceDialog.value) {
+            Dialog(onDismissRequest = { showRollDiceDialog.value = false }) {
+                AlertDialog(
+                    onDismissRequest = { showRollDiceDialog.value = false },
+                    title = { Text("Dice Roll") },
+                    text = {
+                        Column {
+                            // 3 buttons, containing the name of a random location
+                            Button(onClick = { }) {
```
tamaraaPrs, line 181: For maintainability, maybe replace this code copy with something more modular, even if it's only three buttons. 
You can use `repeat` to do that!

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -136,6 +144,73 @@ class MapActivity : ComponentActivity() {
...
+                        Button(onClick = { showRollDiceDialog.value = false }) {
+                            Text("Quit")
+                        }
+                    }
+                )
+            }
+        }
+    }
+
+    private fun rollDice(): com.github.polypoly.app.game.Location {
```
tamaraaPrs, line 202: In our rules we have indicated that when the die is rolled, the game shows us 3 locations. So maybe it's better that this function returns a list of three locations.

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -136,6 +144,73 @@ class MapActivity : ComponentActivity() {
...
+        val diceRollsSum = IntArray(2) { (1..6).random() }.sum() - 1
+
+        // get the 12 closest locations
```
tamaraaPrs, line 205: I would find it better that with each roll of the dice, any place can fall and not the nearest ones. The idea is that when you get a 6 you have to go to the other end of campus. So I suggest this algorithm:
- Sort locations by distance, from closest to least close.
- Remove the one or those who visited the previous round
- Divide the total number of locations by 11 (`n = nbLocations/11`)
- Compute the rest ( `r = nbLocations%11` )
- Draw two random numbers between 1 and 6 and remove two (`diceRollsSum = IntArray(2) { (1..6).random() }.sum() - 2`)
- Select three random locations with an index between `diceRollsSum*(n+1)` and `(diceRollsSum + 1)*(n+1) - 1` if `diceRollsSum < r`, else between `r*(n+1) + (diceRollsSum - r)*n` and `r*(n+1) + (diceRollsSum - r +1)*n -1`
- if there are less than 3 locations in this gap, then take 1 locations before or after

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -136,6 +144,73 @@ class MapActivity : ComponentActivity() {
...
+                            Text("Quit")
+                        }
+                    }
+                )
+            }
+        }
+    }
+
+    private fun rollDice(): com.github.polypoly.app.game.Location {
+        val diceRollsSum = IntArray(2) { (1..6).random() }.sum() - 1
```
tamaraaPrs, line 203: You have to do `- 2` instead of `- 1` because when you add two number between 1 and 6, the range is between 2 and 12, and a tab index begin to 0

tamaraaPrs (changes requested):
Overall the code is fine but for me the algorithm to roll the dice is not quite correct

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
AmLibra: Requires more info about the game, such as previously visited locations, and not sure to agree. To be discussed.
Cannot be implemented yet, read.

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -136,6 +144,73 @@ class MapActivity : ComponentActivity() {
...
+
+        // get the 12 closest locations
+        val closestLocations = markerToLocation.entries
+            .filter { it.value.name != mapViewModel.closeLocation.value?.name } // remove the current location
+            .sortedBy { it.key.position.distanceToAsDouble(mapViewModel.closeLocation.value!!.position) }
+            .take(12)
+
+        return closestLocations[diceRollsSum].value
```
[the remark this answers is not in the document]
AmLibra: Intended feature as in the original game, but done.

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
AmLibra: Done

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
tamaraaPrs: ok ok no problem!

tamaraaPrs (approved):
Good! You can still drop the closest location if you want to be more closed to the original game, but if you drop the closest location, you have to take location since 0 and not 1, as you corrected right there. But, in the original game, the location on which we are is never given because the players would not move otherwise, our players are free to move between each turn.

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -136,6 +144,81 @@ class MapActivity : ComponentActivity() {
...
+    /**
+     * Rolls the dice and returns the location that corresponds to the sum of 2 dice rolls, 3 times
+     * ensuring that the player does not visit the same location twice.
+     */
+    private fun rollDiceLocations(): List<com.github.polypoly.app.game.Location> {
+        val locationsNotToVisitName = mutableListOf(mapViewModel.closeLocation.value?.name)
+
+        val locationsToVisit = mutableListOf<com.github.polypoly.app.game.Location>()
+        for (i in 1..3) {
+            val diceRollsSum = IntArray(2) { (1..6).random() }.sum() - 2
```
SolalPirelli, line 210: This is an interesting choice. I see how it makes sense for fidelity to the original game, but it seems a bit confusing given that you have more dimensions than traditional Monopoly 🤔 looking forward to an explanation in the game rules

PR #88, opened 2023-04-19
Fixing merge issues with the menu

vrooomer:
I noticed that there were some issues with the merge to the main branch. My code was not updated in it so I do this PR to make sure that the main branch is valid.

~ v:)


AmLibra (approved):
Not too much code added, and the added code is good so no issues here

app/src/main/java/com/github/polypoly/app/menu/JoinGameLobbyActivity.kt
```diff
@@ -37,13 +37,14 @@ import com.github.polypoly.app.network.FakeRemoteStorage
...
 import com.github.polypoly.app.ui.theme.PolypolyTheme
+import com.github.polypoly.app.ui.theme.UIElements
 import kotlinx.coroutines.delay
 import timber.log.Timber
 
 /**
  * Activity where the user can join a gameLobby
  */
-class JoinGameLobbyActivity : ComponentActivity() {
+class JoinGameLobbyActivity : MenuActivity("Join Game") {
```
SolalPirelli, line 47: Use string resources instead of hardcoded strings? This won't work if the app is translated in another language, will it?

Issue #79, opened 2023-04-16
Fix Intents init/release in UI tests


Jharaxus:
As of now, initialization and release of intents are done at the beginning and the end of each Android UI test. Therefore, sometimes when a tests fails it doesn't release its intents (as it is supposed to happen after the assertions), which non-deterministically crashes other tests that then look like they double initialize their intents.

Example of such code:
``` kotlin
    @Test
    fun validateProfileButtonGoToTheProfilePage() {

        Intents.init() // here

        // An intent with a name is sent to the activity
        val testIntent =
            Intent(ApplicationProvider.getApplicationContext(), ProfileModifyingActivity::class.java)
        testIntent.putExtra("userId", 1)
        ActivityScenario.launch<ProfileModifyingActivity>(testIntent)

        // Clicking on button
        button.performClick()
        Intents.intended(IntentMatchers.hasComponent(ProfileActivity::class.java.name))

        Intents.release() // here

    }
```

To fix this, one way could be to use the `@Before` and `@After` annotations to initialize and release the intents, therefore avoiding code duplication and enforcing that the release is always done.

PR #89, opened 2023-04-19
Feature/game base

tamaraaPrs:
this PR adds some basics features for the game:
- Improve all the existing system for representing the basic elements of the gameby making them compatible with the database diagram that I've done : https://lucid.app/lucidchart/9118f492-ee01-451e-bd79-d077ba875713/edit?viewport_loc=-403%2C-45%2C2425%2C1178%2C0_0&invitationId=inv_b8f891fe-5a98-4c9c-af7d-710150661d0e (diagram previously inspired by a diagram made by Victor)
- Improve the class `Game` with some additional arguments and function to keep track of the ranking of the players for example
- Create some new basic classes representing basic elements like `GameRules` for the rules, or `PastGame`
- Create a class that represent a `Player` with some function like `earnMoney` or `loseMoney`
- Implement the system of bet to buy a locations with `betToBuy`, the class `LocationBet` and `computeWinningBet`
- Implement the `BonusCard`s and some function about it
- Refractor some packages


app/src/main/java/com/github/polypoly/app/base/PastGame.kt
```diff
@@ -0,0 +1,60 @@
...
+    fun hasWin(userId: Long): Boolean {
+        return usersRank[userId] == 1
+    }
+
+    /**
+     * get the [User] id of the winner
+     * @return the [User] id of the winner
+     */
+    fun getWinnerId(): Long {
+        return usersRank.filterValues { it == 1 }.keys.first()
```
Kaz-ookid, line 33: Here is another example of easy Magic Values to avoid, you could create a `WINNER = 1` constant :)!

Kaz-ookid:
quite a big PR with an impressive amount of work, nice job! 
I only commented what seemed obvious to me, but as you've touched quite all parts of the code, I'll let other people check for problems I could have missed in parts that they've worked on. For my case, I think your PR is all good! It will provide a very good and solid basis to a lot of future tasks in the next sprint :) !

app/src/main/java/com/github/polypoly/app/base/PastGame.kt
tamaraaPrs: In fact the 1 does not represent the winner but the rank of the player. And when you are first, you are the winner. I don't know what you think about it but here I think the 1 makes sense.

app/src/main/java/com/github/polypoly/app/base/game/bonus_card/BonusCard.kt
```diff
@@ -0,0 +1,42 @@
...
+    val description: String,
+) {
+
+    /**
+     * Apply the bonus of the [BonusCard] to the [Player]
+     * @param player the [Player] to apply the bonus to
+     */
+    fun applyBonus(player: Player) {
+        when(id) {
+            0L -> {
```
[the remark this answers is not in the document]
tamaraaPrs: Yeah good Idea, I will create an enum even

app/src/main/java/com/github/polypoly/app/base/game/location/LocationBet.kt
```diff
@@ -0,0 +1,20 @@
...
+ * @property randomNumber The random number generated by the player to know if he/she has won the
+ * bet in the case of a tie
+ * @property timeOfTheBet The time of the bet in Unix time (i.e. the number of milliseconds since
+ * January 1, 1970, 00:00:00 GMT)
+ */
+data class LocationBet (
+    val player: Player = Player(),
+    val amount: Int = 100,
+    val randomNumber: Float = Random.nextFloat(),
+    val timeOfTheBet: Long = System.currentTimeMillis(),
```
vrooomer, line 19: As we discussed together, is it really useful to remember the time of the bet?

app/src/main/java/com/github/polypoly/app/base/game/location/InGameLocation.kt
```diff
@@ -0,0 +1,77 @@
...
+        if(owner != null)
+            throw IllegalStateException("The location is already owned by a player")
+        if(bets.isEmpty())
+            return null
+        val maxAmount = bets.maxOf { it.amount }
+        // case where two players have bet the same amount
+        val allBetsWithMaxAmount = bets.filter { it.amount == maxAmount }
+        val maxRandomNumber = allBetsWithMaxAmount.maxOf { it.randomNumber }
+        // case where two players have bet the same amount and have the same random number
+        val allBetsWithMaxAmountAndSameRdNb = allBetsWithMaxAmount.filter { it.randomNumber == maxRandomNumber }
```
vrooomer, line 68: Maybe such val names are too long?

app/src/main/java/com/github/polypoly/app/base/game/location/InGameLocation.kt
```diff
@@ -0,0 +1,77 @@
...
+ * the location as the name, the base price, the base tax price, etc...
+ * @property level The current level of the location that determine the price and the tax price
+ * @property owner The owner of the location, if the location is not owned, the owner is null
+ * @property bets The list of the bets made on the location by the players in the previous round
+ */
+data class InGameLocation (
+    var location: Location,
+    var level: LocalizationLevel = LocalizationLevel.LEVEL_0,
+    var owner: Player? = null,
+    val bets: List<LocationBet> = listOf()
```
vrooomer, line 17: I don't think keeping the bet list as an argument of the `InGameLocation`. Maybe adding a hashmap in the Game logic itself could be a better idea, as doing a bet is common to all the players and modifying the player state is the Game role.
Tell me if you don't understand me :)

app/src/main/java/com/github/polypoly/app/base/game/rules_and_lobby/GameLobby.kt
```diff
@@ -0,0 +1,97 @@
...
+ * @property admin The admin of the lobby who have created the [GameLobby]
+ * @property rules The rules of the future [Game]
+ * @property name The name of the [GameLobby]
+ * @property code The (secret) code of the [GameLobby]
+ * @property private If the [GameLobby] is private or not
+ */
+class GameLobby(
+    val admin: User = User(),
+    val rules: GameRules = GameRules(),
+    val name: String = "",
```
vrooomer, line 19: Let's be sure that the default name respects what we need (not empty nor blank), so it could be something like "default"

app/src/main/java/com/github/polypoly/app/base/game/rules_and_lobby/GameLobby.kt
```diff
@@ -0,0 +1,97 @@
...
+    val code: String = "",
```
vrooomer, line 20: Same here

app/src/main/java/com/github/polypoly/app/base/game/rules_and_lobby/GameRules.kt
```diff
@@ -0,0 +1,37 @@
...
+ * @property maxRound The maximum number of round before the game end. This settings is available
+ * only in RICHEST_PLAYER mode. Is null if an other game mode is selected.
+ * @property gameMap The map of the game whit the different zones available
+ * @property initialPlayerBalance The initial balance of money of the players
+ */
+data class GameRules (
+    val gameMode: GameMode = GameMode.RICHEST_PLAYER,
+    val minimumNumberOfPlayers: Int = 3,
+    val maximumNumberOfPlayers: Int = 7,
+    val roundDuration: Duration = 2.hours,
```
vrooomer, line 23: As we discussed together, `roundDuration` should better be an integer than a `Duration`. Let's keep it that way because i already changed it in a PR

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -104,8 +107,48 @@ class MapActivity : ComponentActivity() {
...
+                                    bio = "",
+                                    skin = Skin.default(),
+                                    stats = Stats(0,0,0,0,0),
+                                    trophiesWon = listOf(),
+                                    trophiesDisplay = mutableListOf(),
+                                ),
+                                balance = 56,
+                                ownedLocations = listOf(),
+                                roundLost = null,
+                            )),
```
vrooomer, line 151: What is all of that, is it used for testing?

vrooomer (changes requested):
Very nice job overall! I tried to review all your changes even though some code in `MapActivity` and `JoinGameLobby` that aren't you work still look weird.

Let me know if my comments are clear or not!

~ v:)

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
AmLibra: Will be fixed by next week's Great Refactoring of MapActivity! no need to bother Tamara about it !

app/src/main/java/com/github/polypoly/app/base/game/location/LocationBet.kt
tamaraaPrs: Yes because if the two random numbers are equal, we use the time of the bet to decide between the two players, so that we are sure that all the players have the same winner

app/src/main/java/com/github/polypoly/app/base/game/location/InGameLocation.kt
tamaraaPrs: I don't agree, it modularizes better because we don't just have a big Game class that does everything like that. The only thing I should have done though is not to put it in the constructor, I'm going to fix that right away.

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
tamaraaPrs: I only changed this part to make it compile

app/src/main/java/com/github/polypoly/app/base/game/rules_and_lobby/GameLobby.kt
tamaraaPrs: done

app/src/main/java/com/github/polypoly/app/base/game/rules_and_lobby/GameLobby.kt
tamaraaPrs: done

app/src/main/java/com/github/polypoly/app/base/game/rules_and_lobby/GameRules.kt
tamaraaPrs: done

app/src/main/java/com/github/polypoly/app/base/game/location/InGameLocation.kt
tamaraaPrs: I reduced the name of the variables a little bit while leaving them understandable

app/src/main/java/com/github/polypoly/app/base/game/location/LocationBet.kt
vrooomer: We discussed it already and we agreed that this way is a bit "overkill" as the probability to have the same bet AND the same random number is very small, let's always say that the bet number 1 (stored in the DB) wins. Having a time for each bet might be too much, what do you think?

vrooomer: _After a discussion I agree that the time can be stored as we're not sure that the order is the DB is deterministic_

vrooomer (approved):

app/src/androidTest/java/com/github/polypoly/app/base/game/PlayerTest.kt
```diff
@@ -0,0 +1,237 @@
...
+        val testPlayer = Player(testUser1, 100, listOf())
+        testPlayer.loseMoney(50)
+        assertEquals(50, testPlayer.getBalance())
+    }
+
+    @Test
+    fun loseMoneyDoesNotDecreaseTheMoneyOfThePlayerBelowZero() {
+        val testPlayer = Player(testUser1, 100, listOf())
+        testPlayer.loseMoney(150)
+        assertEquals(0, testPlayer.getBalance())
```
Jharaxus, line 86: Same as above

app/src/androidTest/java/com/github/polypoly/app/base/game/PlayerTest.kt
```diff
@@ -0,0 +1,237 @@
...
+
+    @Test
+    fun loseMoneyThrowsAnExceptionIfTheAmountIsNegative() {
+        val testPlayer = Player(testUser1, 100, listOf())
+        try {
+            testPlayer.loseMoney(-50)
+            fail("Should have thrown an exception")
+        } catch (e: IllegalArgumentException) {
+            assertEquals("The amount of money lost cannot be negative or zero", e.message)
+        }
```
Jharaxus, line 97: You should use assertThrows from junit instead of try/catch

Jharaxus (changes requested):

app/src/androidTest/java/com/github/polypoly/app/base/game/PlayerTest.kt
```diff
@@ -0,0 +1,237 @@
...
+    @Test
+    fun compareToConsiderEqualThePlayerWhoLostAtTheSameTime() {
+        assertTrue(testPlayer6.compareTo(testPlayer5) == 0)
+    }
+
+    @Test
+    fun loseMoneyDecreaseTheMoneyOfThePlayer() {
+        val testPlayer = Player(testUser1, 100, listOf())
+        testPlayer.loseMoney(50)
+        assertEquals(50, testPlayer.getBalance())
```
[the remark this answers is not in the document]
Jharaxus: This applies to several other tests below (I did not comment all of them)

app/src/androidTest/java/com/github/polypoly/app/base/game/GameTest.kt
```diff
@@ -0,0 +1,93 @@
...
+    private val testUser4 = User(42042045, "test_user4", "", emptySkin, zeroStats, listOf(), mutableListOf())
+    private val testUser5 = User(42042046, "test_user5", "", emptySkin, zeroStats, listOf(), mutableListOf())
+    private val testUser6 = User(42042047, "test_user6", "", emptySkin, zeroStats, listOf(), mutableListOf())
+    private val testDuration = 2
+    private val gameRules = GameRules(GameMode.RICHEST_PLAYER, 3, 7,
+        testDuration, 10, LocationRepository.getZones(), 200)
+    private val gameLobby = GameLobby(testUser1, gameRules, "test_game", "123456", false)
+
+    @Before
+    fun startIntents() {
```
Jharaxus, line 34: Why is it called `startIntents` if it doesn't start intents? (it is also the case in below tests)
Actually such a function can be left out as it should be soon solved through issue [#79](https://github.com/polypoly-team/android-app/issues/79)

app/src/androidTest/java/com/github/polypoly/app/base/game/GameTest.kt
```diff
@@ -0,0 +1,93 @@
...
+        gameLobby.addUser(testUser2)
+        gameLobby.addUser(testUser3)
+        gameLobby.addUser(testUser4)
+        gameLobby.addUser(testUser5)
+        gameLobby.addUser(testUser6)
```
Jharaxus, line 39: You can inherit from the newly created PolyPolyTest which does this for you.

app/src/androidTest/java/com/github/polypoly/app/network/RemoteDBTest.kt
```diff
@@ -1,49 +1,92 @@
...
-
-    private fun <T : Any> classCanBeStoredInDB(clazz: KClass<T>) {
-        val noArgsKey = "no-args-obj"
-        val noArgsConstructor = clazz.java.getDeclaredConstructor()
-        val noArgsInstance = noArgsConstructor.newInstance()
-        addDataToDB(listOf(noArgsInstance), listOf(noArgsKey))
-        assertEquals(
-            noArgsInstance,
-            remoteDB.getValue(noArgsKey, clazz).get(TIMEOUT_DURATION, TimeUnit.SECONDS)
-        )
```
Jharaxus, line 27: I think something went wrong with the merge, this should be kept and the code below (marked added) is outdated.

Jharaxus: Okay it actually looks like you rolled back the file back to its previous version during merging so something definitely went wrong.

app/src/main/java/com/github/polypoly/app/base/PastGame.kt
```diff
@@ -0,0 +1,60 @@
...
+    val date: Long = 0,
+    val duration: Long = 0,
+) {
+
+    /**
+     * Return if the [User] has won the [Game]
+     * @param userId the id of the [User]
+     * @return true if the user has won the [Game], false otherwise
+     */
+    fun hasWin(userId: Long): Boolean {
```
Jharaxus, line 24: Little typo: it should be `hasWon` (as in the comments)

app/src/main/java/com/github/polypoly/app/base/PastGame.kt
```diff
@@ -0,0 +1,60 @@
...
+        return users
+    }
+
+    /**
+     * get the rank of the [User]
+     * @param userId the id of the [User]
+     * @return the rank of the [User] or 0 if the [User] is not in the [PastGame]
+     */
+    fun getRank(userId: Long): Int {
+        return usersRank[userId] ?: 0
```
Jharaxus, line 50: I think we should throw an exception instead of returning 0 as this is definitely an access error so we should clearly notified that the request was invalid: I believe the caller shouldn't have to worry about the fact that it can fail = return 0 here, as only valid userIds should be provided, so an exception makes sense as it wouldn't need to be explicitly handled as it would really be an error to make such wrong call.

app/src/main/java/com/github/polypoly/app/base/game/Game.kt
```diff
@@ -0,0 +1,158 @@
...
+ * (seconds since 1970-01-01T00:00:00Z)
+ */
+class Game private constructor(
+    val admin: User = User(),
+    var players: List<Player> = listOf(),
+    val rules: GameRules = GameRules(),
+    val dateBegin: Long = System.currentTimeMillis(),
+) {
+
+    private val inGameLocations: List<InGameLocation> = rules.gameMap.flatMap { it.locations.map { location ->
```
Jharaxus, line 28: `it` shouldn't be used as is in lambdas (even if the fact that it is implicitly define is useful), but should instead be renamed with a more expressive name: `.flatMap { good_name ->  good_name.locations...`

app/src/main/java/com/github/polypoly/app/base/game/Game.kt
```diff
@@ -0,0 +1,158 @@
...
+     * @return true if the game is finished, false otherwise
+     * @throws IllegalStateException if the game mode is RICHEST_PLAYER and maxRound is null
+     */
+    private fun isGameFinished(): Boolean {
+        return when(rules.gameMode) {
+            GameMode.LAST_STANDING -> {
+                players.filter { !it.hasLose() }.size <= 1
+            }
+            GameMode.RICHEST_PLAYER -> {
+                if(rules.maxRound == null)
```
Jharaxus, line 58: This actually does not need to be tested (even if the attribute is kept null which I don't recommend as pointed out in another comment). Indeed, the object (Game) is supposed to always be well defined / well formed. So it's the constructor's job to check that property once, and after that any function modifying the object's internal state should keep its consistency so that such test is never needed.

app/src/main/java/com/github/polypoly/app/base/game/Game.kt
```diff
@@ -0,0 +1,158 @@
...
+     * @return a [Map] of the [Player]s' [User] id and their rank
+     */
+    fun ranking(): Map<Long, Int> {
+        val playersSorted = players.sortedDescending()
+        val map = playersSorted.mapIndexed { index, player -> player.user.id to index+1 }.toMap().toMutableMap()
+        for(i in 1 until (players.size)) {
+            if(playersSorted[i].compareTo(playersSorted[i-1]) == 0) {
+                map[playersSorted[i-1].user.id]?.let{ map[playersSorted[i].user.id] = it }
+            }
+        }
```
Jharaxus, line 76: This could be refactor using two iterator variable `currentPlayer` and `previousPlayer` instead of using several times `playersSorted[i]` and `playersSorted[i-1]`
Moreover, I believe a simple affectation in the map would be clearer than the `let` + lambda

app/src/main/java/com/github/polypoly/app/base/game/Game.kt
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * End the game and return a PastGame object
+     * @return the PastGame object
+     * @throws IllegalStateException if the game is not finished
+     */
+    fun endGame(): PastGame {
+        if(!isGameFinished()) throw IllegalStateException("can't end the game now")
+        return PastGame(
+            users = players.map { it.user },
```
Jharaxus, line 88: As above you should not use `it` (it is also the case below). Moreover, such call can be replaced by `players.map(Player::user)` which is more meaningful.

Jharaxus (commented):
Several `it` can be replaced with names with more intent.

app/src/main/java/com/github/polypoly/app/base/game/rules_and_lobby/GameRules.kt
```diff
@@ -0,0 +1,37 @@
...
+ * only in RICHEST_PLAYER mode. Is null if an other game mode is selected.
+ * @property gameMap The map of the game whit the different zones available
+ * @property initialPlayerBalance The initial balance of money of the players
+ */
+data class GameRules (
+    val gameMode: GameMode = GameMode.RICHEST_PLAYER,
+    val minimumNumberOfPlayers: Int = 3,
+    val maximumNumberOfPlayers: Int = 7,
+    val roundDuration: Int = 0,
+    val maxRound: Int? = null,
```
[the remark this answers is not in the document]
tamaraaPrs: I don't agree, for me if the nullable variables make sense to be nullable we should keep them. Not setting this variable to null when you don't need to use it is taking the risk of using it unintentionally and no warning is indicated. What we must limit as much as possible are the `!!` which are a bad practice, the code must be a maximum null safety. Besides there is a list of good practices that allow to avoid `!!` by making the code null safe like for example assigning the nullable attribute to a local variable at the start of the function and test if the local variable is null. Normally in this case the code compiles without `!!`.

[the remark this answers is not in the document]
tamaraaPrs: This is the page for Kotlin on null safety: https://kotlinlang.org/docs/null-safety.html

app/src/androidTest/java/com/github/polypoly/app/base/game/GameTest.kt
```diff
@@ -0,0 +1,93 @@
...
+    private val emptySkin = Skin(0,0,0)
+    private val zeroStats = Stats(0,0,0,0,0)
+    private val testUser1 = User(42042042, "test_user1", "", emptySkin, zeroStats, listOf(), mutableListOf())
+    private val testUser2 = User(42042043, "test_user2", "", emptySkin, zeroStats, listOf(), mutableListOf())
+    private val testUser3 = User(42042044, "test_user3", "", emptySkin, zeroStats, listOf(), mutableListOf())
+    private val testUser4 = User(42042045, "test_user4", "", emptySkin, zeroStats, listOf(), mutableListOf())
+    private val testUser5 = User(42042046, "test_user5", "", emptySkin, zeroStats, listOf(), mutableListOf())
+    private val testUser6 = User(42042047, "test_user6", "", emptySkin, zeroStats, listOf(), mutableListOf())
+    private val testDuration = 2
+    private val gameRules = GameRules(GameMode.RICHEST_PLAYER, 3, 7,
```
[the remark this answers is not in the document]
tamaraaPrs: fix done

app/src/androidTest/java/com/github/polypoly/app/base/game/PlayerTest.kt
```diff
@@ -0,0 +1,237 @@
...
+        gameLobby.addUser(testUser6)
+        gameLobby.start()
+    }
+
+
+    @Test
+    fun compareToConsiderGreaterThePlayerWithMoreMoney() {
+       assertTrue(testPlayer1 < testPlayer2)
+       assertTrue(testPlayer2 > testPlayer1)
+    }
```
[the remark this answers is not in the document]
tamaraaPrs: fix done

app/src/androidTest/java/com/github/polypoly/app/base/game/PlayerTest.kt
```diff
@@ -0,0 +1,237 @@
...
+class PlayerTest {
+
+    private val emptySkin = Skin(0,0,0)
+    private val zeroStats = Stats(0,0,0,0,0)
+    private val testUser1 = User(42042042, "test_user1", "", emptySkin, zeroStats, listOf(), mutableListOf())
+    private val testUser2 = User(42042043, "test_user2", "", emptySkin, zeroStats, listOf(), mutableListOf())
+    private val testUser3 = User(42042044, "test_user3", "", emptySkin, zeroStats, listOf(), mutableListOf())
+    private val testUser4 = User(42042045, "test_user4", "", emptySkin, zeroStats, listOf(), mutableListOf())
+    private val testUser5 = User(42042046, "test_user5", "", emptySkin, zeroStats, listOf(), mutableListOf())
+    private val testUser6 = User(42042047, "test_user6", "", emptySkin, zeroStats, listOf(), mutableListOf())
```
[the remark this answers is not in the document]
tamaraaPrs: fix done

app/src/androidTest/java/com/github/polypoly/app/base/game/PlayerTest.kt
```diff
@@ -0,0 +1,237 @@
...
+    private val testPlayer1 = Player(testUser1, 100, listOf())
+    private val testPlayer2 = Player(testUser2, 200, listOf())
+    private val testPlayer3 = Player(testUser3, 100, listOf())
+    private val testPlayer4 = Player(testUser4, 0, listOf(), 4)
+    private val testPlayer5 = Player(testUser5, 0, listOf(), 5)
+    private val testPlayer6 = Player(testUser6, 0, listOf(), 5)
```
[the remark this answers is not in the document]
tamaraaPrs: fix done

app/src/androidTest/java/com/github/polypoly/app/base/game/PlayerTest.kt
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    @Test
+    fun compareToConsiderEqualThePlayerWithSameMoney() {
+        assertTrue(testPlayer1.compareTo(testPlayer3) == 0)
+    }
+
+    @Test
+    fun compareToConsiderSmallerThePlayerWhoLostFirst() {
+        assertTrue(testPlayer4 < testPlayer5)
```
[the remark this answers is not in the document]
tamaraaPrs: Name fix, I'm not agree with the rest of the comment (You could argue that for any test)

app/src/androidTest/java/com/github/polypoly/app/base/game/PlayerTest.kt
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    @Test
+    fun compareToConsiderEqualThePlayerWhoLostAtTheSameTime() {
```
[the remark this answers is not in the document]
tamaraaPrs: To know better where the problem comes from when the test fails, it is not advised in the Software Engineering course to put several tests in one

app/src/androidTest/java/com/github/polypoly/app/base/game/PlayerTest.kt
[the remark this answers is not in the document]
tamaraaPrs: fix done

app/src/androidTest/java/com/github/polypoly/app/base/game/PlayerTest.kt
tamaraaPrs: fix done

app/src/androidTest/java/com/github/polypoly/app/base/game/PlayerTest.kt
tamaraaPrs: fix done

app/src/androidTest/java/com/github/polypoly/app/base/game/GameTest.kt
tamaraaPrs: my bad for the name. It is fix. I don't need to start the Intents anyway

app/src/androidTest/java/com/github/polypoly/app/base/game/GameTest.kt
tamaraaPrs: I want to create my own game specific to my test

app/src/androidTest/java/com/github/polypoly/app/network/RemoteDBTest.kt
tamaraaPrs: ok I copy your `RemoteDBTest` from the main

app/src/main/java/com/github/polypoly/app/base/PastGame.kt
tamaraaPrs: thank you, fix done

app/src/main/java/com/github/polypoly/app/base/PastGame.kt
tamaraaPrs: thank you, fix done

app/src/main/java/com/github/polypoly/app/base/game/Game.kt
tamaraaPrs: thank you, fix done

app/src/main/java/com/github/polypoly/app/base/game/Game.kt
tamaraaPrs: If I don't check if the max round is null, I have to use `!!` and it's not null safety, even if this senario should not happen, a clear exception is thrown when the error occurs

app/src/main/java/com/github/polypoly/app/base/game/Game.kt
tamaraaPrs: thank you, fix done

app/src/main/java/com/github/polypoly/app/base/game/Game.kt
tamaraaPrs: thank you, fix done

app/src/androidTest/java/com/github/polypoly/app/base/game/PlayerTest.kt
```diff
@@ -0,0 +1,245 @@
...
+
+    @Test
+    fun loseMoneyThrowsAnExceptionIfTheAmountIsNegative() {
+        val testPlayer = Player(TEST_USER_0, 100, listOf())
+        val thrown = assertThrows(IllegalArgumentException::class.java) {
+            testPlayer.loseMoney(-50)
+        }
+        assertNotNull(thrown.message)
+        val message = thrown.message
+        if (message != null) {
```
SolalPirelli, line 79: this has to be true given the assertNotNull earlier, right?

app/src/androidTest/java/com/github/polypoly/app/base/game/PlayerTest.kt
```diff
@@ -0,0 +1,245 @@
...
+            assertEquals("The location is already owned by someone", message)
+        }
+    }
+
+    @Test
+    fun betToBuyThrowsAnExceptionIfTheAmountIsSmallerThanTheLocationPrice() {
+        val testPlayer = Player(TEST_USER_0, 100, listOf())
+        val thrown = assertThrows(IllegalArgumentException::class.java) {
+            testPlayer.betToBuy(InGameLocation(LocationRepository.getZones()[0].locations[0]), 100)
```
SolalPirelli, line 218: IMHO it'd be clearer to create your own fake location for such tests, since it's not clear when reading this why `100` wouldn't be enough, and any change to the real data might make this test fail even though it's not related to the real data

Jharaxus (approved):

app/src/main/java/com/github/polypoly/app/base/game/Game.kt
```diff
@@ -0,0 +1,160 @@
...
+     */
+    fun getPlayer(userId: Long): Player? {
+        return players.find { it.user.id == userId }
+    }
+
+    /**
+     * Compute the winner of the bets, notify the players and update the balance of the players
+     * and the location owner.
+     */
+    fun computeAllWinnersOfBets() {
```
SolalPirelli, line 119: Terminology-wise, I think you mean `bid`, which is the thing people do at auctions, rather than `bet`, which is a form of gambling

app/src/main/java/com/github/polypoly/app/base/game/Player.kt
```diff
@@ -0,0 +1,167 @@
...
+    private var balance: Int = 0,
+    private var ownedLocations: List<InGameLocation> = listOf(),
+    private var roundLost: Int? = null,
+) : Comparable<Player> {
+
+    /**
+     * If the player has lost the game (i.e. if the player has no more money)
+     * @return true if the player has lost the game, false otherwise
+     */
+    fun hasLose(): Boolean {
```
SolalPirelli, line 28: spelling: `hasLost` (like the doc comment correctly states already)

app/src/main/java/com/github/polypoly/app/base/game/bonus_card/BonusCard.kt
```diff
@@ -0,0 +1,35 @@
...
+     */
+    fun applyBonus(player: Player) {
+        when(this) {
+            TEACHING_ASSISTANT_PAYDAY -> {
+                player.earnMoney(100)
+            }
+            TUITION_FEES_TO_PAY -> {
+                player.loseMoney(100)
+            }
+            // TODO : implement other bonus cards
```
SolalPirelli, line 32: looking forward to reading those

app/src/main/java/com/github/polypoly/app/base/game/bonus_card/BonusCard.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    TEACHING_ASSISTANT_PAYDAY(0, title = "Teaching assistant payday", description = "You finally got your student assistant salary you've been waiting for after spending hours helping BA1s debug their code. Collect 100CHF!"),
+    TUITION_FEES_TO_PAY(1, title = "Tuition fees to pay!", description = "It's time to pay your tuition fees: you have passed the is academia deadline for far too long... Lose 100CHF.");
+
+    /**
+     * Apply the bonus of the [BonusCard] to the [Player]
+     * @param player the [Player] to apply the bonus to
+     */
+    fun applyBonus(player: Player) {
+        when(this) {
```
SolalPirelli, line 25: it might make more sense for each bonus card to contain its own action, rather than split the definition of each card and the definition of which action each card does? As it stands this file mixes functional and object-oriented programming in  a bit of an odd way

app/src/main/java/com/github/polypoly/app/base/user/Skin.kt
```diff
@@ -0,0 +1,59 @@
...
+        }
+        // Checks that the skin exists
+        testIdRange(idHead, NB_HEADS)
+        testIdRange(idBody, NB_BODIES)
+        testIdRange(idLegs, NB_LEGS)
+    }
+
+    companion object {
+        // TODO: Replace this by a list of skin parts
+        // The number of possible heads, bodies and legs
```
SolalPirelli, line 30: polypoly: cannibal edition 😱

app/src/main/java/com/github/polypoly/app/base/game/bonus_card/BonusCard.kt
tamaraaPrs: Yes, it clearly makes sense. I'm going to fix this

app/src/main/java/com/github/polypoly/app/base/game/Player.kt
tamaraaPrs: thank you, fix done

app/src/main/java/com/github/polypoly/app/base/game/Game.kt
tamaraaPrs: Thank you for this terminology tip, I didn't know it! Fix done

app/src/androidTest/java/com/github/polypoly/app/base/game/PlayerTest.kt
tamaraaPrs: Yes my bad, when I changed the tests I forgot to remove this part. Fix done

app/src/androidTest/java/com/github/polypoly/app/base/game/PlayerTest.kt
tamaraaPrs: Yes, I'm agree. Fix done

app/src/androidTest/java/com/github/polypoly/app/network/RemoteDBTest.kt
```diff
@@ -28,7 +28,7 @@ class RemoteDBTest: PolyPolyTest(true, false) {
     }
 
     @Test
-    fun gameLobbyCanBeStoredInDB() {
+    fun gameLobbyCanBççeStoredInDB() {
```
AlSchlo, line 31: Typo??

PR #90, opened 2023-04-20
Update google-services.json

Kaz-ookid:
added every team member's SHA1 key to firebase. Should not need to be updated anymore after that :)!


jadtl (approved):
Looks good to me, thanks :)

PR #91, opened 2023-04-20
Game Lobby Activity (dummy version)

vrooomer:
_This feature isn't done but I open a draft PR so y'all can see the new code before the end of the sprint I didn't code that much lines, that mostly comes from the branch of Tamara. When this sprint is realy i'll pull everything so the changes will correspond to my code only._

The main goal of this feature is to be able to create, modify, access and delete `GameLobby`s to the DB. In that way, the `GameLobbyActivity` receives a game lobby code when started, so that it can get the corresponding lobby from the DB. 
Also, in this activity you will be able to see the players in the lobby in real time!

When enough players are in the lobby, the admin can then start the game which will start a `Game` and destroy the `GameLobby`.

Finally, if the user quits the lobby, it is simply destroyed and the other players should be kicked out.

=========================== EDIT ==============================

The feature isn't done but we need a way to access to `MapActivity` ASAP so I added a button in the activity for that purpose.
I had to delete a lot of work related to the DB because we won't be working that way and I'll finish that task on a further PR.

~ v:)


SolalPirelli:
You can make this kind of PR clearer by making it a PR against the original branch it's modifying rather than main. This can be done (and reverted) after creating the PR, so you could even switch it back to being against main after the original branch has been merged.

app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
```diff
@@ -51,6 +51,14 @@ open class RemoteDB(
         return valuePromise
     }
 
+    override fun getSnapshot(key: String): CompletableFuture<DataSnapshot> {
+        return getValueAndThen(key, { data, valuePromise ->
+            valuePromise.complete(data)
+        }, { valuePromise ->
+            valuePromise.completeExceptionally(IllegalAccessException("No value found for key <$key>"))
```
SolalPirelli, line 58: same remark as in another PR, this exception is meant for illegal reflection accesses to class members, NoSuchElementException would probable be a better fit, or maybe create your own exception type

app/src/main/java/com/github/polypoly/app/base/game/rules_and_lobby/kotlin/GameLobby.kt
```diff
@@ -1,7 +1,8 @@
-package com.github.polypoly.app.base.game.rules_and_lobby
+package com.github.polypoly.app.base.game.rules_and_lobby.kotlin
```
SolalPirelli, line 1: Why include `kotlin` in the package? Will there be another programming language in the codebase that won't use this?

app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
vrooomer: I agree with that, this is not my code but i'll mention @Jharaxus on that

app/src/main/java/com/github/polypoly/app/base/game/rules_and_lobby/kotlin/GameLobby.kt
vrooomer: This isn't the cleanest way to do it but we create `kotlin` packages when we want to separate back/frontend

app/src/main/java/com/github/polypoly/app/base/game/rules_and_lobby/GameLobbyActivity.kt
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+    }
+
+    // ===================================================== GAME LOBBY COMPONENTS
+
+    @Composable
+    private fun PlayerGrid() {
+        val numberOfPlayers = gameLobby.value.usersRegistered.size
+        val numberOfRows: Int = numberOfPlayers / playerPerRow
```
tamaraaPrs, line 94: What happens if the division result is not an exact integer? One row less will be counted because of the rounding! The integer division in kotlin is the same of java.

app/src/main/java/com/github/polypoly/app/base/game/rules_and_lobby/GameLobbyActivity.kt
```diff
@@ -0,0 +1,193 @@
...
+                    if(i != numberOfRows - 1) {
+                        Spacer(modifier = Modifier.height(Padding.medium))
+                    }
+                }
+            }
+        }
+    }
+
+    @Composable
+    private fun PlayerIcon(user: User?) {
```
tamaraaPrs, line 132: Same for the documentation :)

app/src/main/java/com/github/polypoly/app/base/game/rules_and_lobby/GameLobbyActivity.kt
```diff
@@ -0,0 +1,193 @@
...
+                    .size(30.dp)
+            )
+            if(user != null) {
+                Text(text = user.name)
+            }
+        }
+    }
+
+    /**
+     * TODO: dummy button that redirects to MapActivity
```
tamaraaPrs, line 155: I find the description of this TODO confusing because it sounds like a dummy button that redirects to MapActivity, while this is what the code below seems to do. What did you mean by this TODO?

app/src/main/java/com/github/polypoly/app/base/game/rules_and_lobby/kotlin/GameRules.kt
```diff
@@ -26,12 +24,14 @@ data class GameRules (
     val initialPlayerBalance: Int = 500,
 ) {
 
+    private val maxRoundHours = Int.MAX_VALUE
```
tamaraaPrs, line 27: It's interesting to restrict the time that can be put in, it's just that it's not useful to initialize it to `Int.MAX_VALUE`, but I guess you just did that while waiting for the group to discuss it. Maybe put a `TODO` so we don't forget to change it!

tamaraaPrs (changes requested):
Great work, just pay attention to the documentation of your code and I've made some remarks about edge cases that you can take into account for the purpose of responsive layout and resilience of the code. 
Don't hesitate to go over your TODOs, I wonder if you don't forget to remove some of them after you've done them! 

Thank you for your efficient work and your reactivity on this PR which will be very useful to the whole team! 😃

tamaraaPrs (changes requested):
Don't forget to add some tests too!

app/src/main/java/com/github/polypoly/app/base/game/rules_and_lobby/GameLobbyActivity.kt
```diff
@@ -0,0 +1,193 @@
...
+        receiveGameLobby()
+        setContent { GameLobbyContent() }
+    }
+
+    @Preview(showBackground = true)
+    @Composable
+    private fun GameLobbyPreview() { GameLobbyContent() }
+
+    // ===================================================== MAIN CONTENT
+    @Composable
```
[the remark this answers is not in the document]
vrooomer: I documented now!

vrooomer:
I changed what you advised, now it should be cleaner!

As this PR only implements a dummy version of the lobby, most of the code will change in the future so I decided to only create one dummy test that checks if the "go" button launches the wanted activity.

Note that this is really not a final version and that the only reason that I do this PR is because we need a way to go to the Game/MapActivity ASAP.

~ v:)

app/src/main/java/com/github/polypoly/app/base/game/rules_and_lobby/GameLobbyActivity.kt
tamaraaPrs: it works :)

app/src/main/java/com/github/polypoly/app/base/game/rules_and_lobby/kotlin/GameLobby.kt
SolalPirelli: you could call them backend and frontend then :-)

PR #92, opened 2023-04-20
Add a game menu to the map's heads-up display

jadtl:
This PR adds a game menu to the map's heads-up display, in the form of a drop-down button that displays a row of menu buttons right above it with a slide-in animation.

I had to put this menu on the left of the screen instead of the player's information that I moved to the bottom right of the activity because when the menu is in the right, opening it causes triggers a re-layout of the box that encloses the drop-down button and the menu buttons, causing the drop-down button to instantenously jump on the left.

I added more documentation to the HUD.

### DIFF
--- a/app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
+++ b/app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
@@ -58,6 +58,7 @@ import com.github.polypoly.app.BuildConfig
 import com.github.polypoly.app.R
 import com.github.polypoly.app.game.PlayerGlobalData
 import com.github.polypoly.app.map.LocationRepository.getZones
+import com.github.polypoly.app.menu.MenuComposable
 import com.github.polypoly.app.ui.theme.PolypolyTheme
 import com.github.polypoly.app.ui.theme.Shapes
 import com.github.polypoly.app.utils.Padding
@@ -508,6 +509,7 @@ class MapActivity : ComponentActivity() {
         HudPlayer(playerData)
         HudOtherPlayersAndGame(otherPlayersData, round)
         HudLocation(location = mapViewModel.closeLocation.value?.name ?: "")
+        HudGameMenu()
     }
 
     /**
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -528,7 +530,8 @@ class MapActivity : ComponentActivity() {
     }
 
     /**
-     * The HUD for the player stats
+     * The HUD for the player stats, it displays basic information such as their balance,
+     * and a button shows complete information on click
```
tamaraaPrs, line 534: you can precise the description of the param in the documentation of the function with `@param`!

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -574,7 +577,9 @@ class MapActivity : ComponentActivity() {
     }
 
     /**
-     * The HUD that shows the stats for other players and the game
+     * The HUD that shows the stats for other players and the game, it's a button on the top left
+     * that expands and collapses a tab that contains information about the game and the other
+     * players
```
tamaraaPrs, line 582: same

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -621,7 +628,8 @@ class MapActivity : ComponentActivity() {
     }
 
     /**
-     * The HUD for the game stats
+     * The HUD for the game stats, it displays basic information such as the current round,
+     * and a button shows complete information on click
```
tamaraaPrs, line 632: same

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
```diff
@@ -658,7 +666,8 @@ class MapActivity : ComponentActivity() {
     }
 
     /**
-     * The HUD for the stats of other players
+     * The HUD for the stats of other players, it displays basic information such as their balance,
+     * and a button shows complete information on click
```
tamaraaPrs, line 670: same

tamaraaPrs (approved):
Apart from the detail of the comments, everything seems fine to me!

app/src/main/java/com/github/polypoly/app/map/MapActivity.kt
jadtl: Thanks for the suggestion!

SolalPirelli: fyi, "préciser" en français -> "to specify" in English, precise is only an adjective ^^

PR #106, opened 2023-04-24
Fix: valid main after merge & CONVENTIONS

vrooomer:
After the merge of this sprint, some functionalities were removed by accident so I do this PR to fix that issue.

Also, I deleted the music that was added to `SigningActivity` so that it begins only once at `WelcomeActivity`

As the code is a bit messy (not using the same conventions everywhere), I also created a file CONVENTIONS.md so that we all use the code in a same fashion, of course feel to complete it when you think it is necessary

~ v:)


Kaz-ookid (approved):
thank you for taking the time to fix these!
These are required and important changes to merge before starting any new work. Nice job :)

PR #112, opened 2023-04-25
Add MockDB instead of using real DB for tests

Jharaxus:
Using the real database for testing purpose may lead to side effects / inconsistent behaviors.
This PR adds the class MockDB as a mock version of RemoteDB, aiming at being used in all tests instead (apart from RemoteDB tests ofc).

### DIFF
--- a/app/src/androidTest/java/com/github/polypoly/app/commons/MockDB.kt
+++ b/app/src/androidTest/java/com/github/polypoly/app/commons/MockDB.kt
@@ -4,6 +4,9 @@ import com.github.polypoly.app.network.IRemoteStorage
 import java.util.concurrent.CompletableFuture
 import kotlin.reflect.KClass
 
+/**
+ * Mock version of RemoteDB for testing purpose
+ */
 class MockDB: IRemoteStorage {
 
     private val keysHierarchy: MutableMap<String, MutableList<String>> = mutableMapOf()
@@ -58,15 +61,16 @@ class MockDB: IRemoteStorage {
         for (index in 0 until hierarchy.size - 1) {
             val parent = hierarchy[index]
             val child = hierarchy[index + 1]
-            if (!keysHierarchy.containsKey(parent)) {
-                keysHierarchy[parent] = mutableListOf()
-            }
+            keysHierarchy.putIfAbsent(parent, mutableListOf())
             keysHierarchy[parent]!!.add(child)
         }
         data[keyCleaned] = value as Any
         return CompletableFuture.completedFuture(true)
     }
 
+    /**
+     * Clears all value in the mock database
+     */
     fun clear() {
         data.clear()
         keysHierarchy.clear()
--- a/app/src/androidTest/java/com/github/polypoly/app/commons/MockDB.kt
+++ b/app/src/androidTest/java/com/github/polypoly/app/commons/MockDB.kt
@@ -1,0 +1,74 @@
+package com.github.polypoly.app.commons
+
+import com.github.polypoly.app.network.IRemoteStorage
+import java.util.concurrent.CompletableFuture
+import kotlin.reflect.KClass
+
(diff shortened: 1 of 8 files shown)


app/src/androidTest/java/com/github/polypoly/app/commons/MockDB.kt
```diff
@@ -0,0 +1,74 @@
...
+        @Suppress("UNCHECKED_CAST")
+        return CompletableFuture.completedFuture(data[cleanKey(key)] as T)
+    }
+
+    override fun <T : Any> getAllValues(key: String, clazz: KClass<T>): CompletableFuture<List<T>> {
+        val objs = mutableListOf<T>()
+        val parentKeyCleaned = cleanKey(key)
+        @Suppress("UNCHECKED_CAST")
+        for (child in keysHierarchy[parentKeyCleaned] ?: listOf())
+            objs.add(data["$parentKeyCleaned/$child"] as T)
```
tamaraaPrs, line 29: You can use safe cast instead :
`(data["$parentKeyCleaned/$child" as? T)?.let { objs.add(it) }` (i.e. if the cast succeeds and the key exist, add it, else do nothing)

tamaraaPrs (changes requested):
Great work! Be careful to document your code though. I also made some remarks about null safety. Thank you so much for this PR that you did super fast, it will unlock us quite a bit for the CI 😅

app/src/androidTest/java/com/github/polypoly/app/commons/MockDB.kt
```diff
@@ -0,0 +1,74 @@
...
+    override fun <T> setValue(key: String, value: T): CompletableFuture<Boolean> {
+        val keyCleaned = cleanKey(key)
+        val hierarchy = keyCleaned.split("/")
+        for (index in 0 until hierarchy.size - 1) {
+            val parent = hierarchy[index]
+            val child = hierarchy[index + 1]
+            if (!keysHierarchy.containsKey(parent)) {
+                keysHierarchy[parent] = mutableListOf()
+            }
+            keysHierarchy[parent]!!.add(child)
```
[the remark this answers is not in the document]
Jharaxus: Thanks for the explanation, I used `computeIfPresent` instead of the `if (!keysHierarchy.containsKey(parent)) {...}`. However I kept the `!!` as even though it usually is a bad practice:
- Here it should be safe to use it as a missing value would be added the line above if it didn't exist yet
- For the database's consistency, the child must always be added or the structure would be ill-formed. Therefore, throwing an exception if somehow it fails seems legit to me, especially considering that this shouldn't happen thanks to the point above.

app/src/androidTest/java/com/github/polypoly/app/commons/MockDB.kt
Jharaxus: This was actually an intended feature: this class is designed for test purposes, not for live application deployment. Therefore, if somehow an incorrect value is put in the DB which results in a failure when retrieving it (either through `getValue` or `getAllValues`), this means either the test or the code tested had issues that led to an incorrect DB state, so it should crash to reflect this situation (through the exception thrown by the unsafe cast).

app/src/androidTest/java/com/github/polypoly/app/commons/MockDB.kt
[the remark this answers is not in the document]
tamaraaPrs: yes, I understand!

app/src/androidTest/java/com/github/polypoly/app/commons/MockDB.kt
tamaraaPrs: ok I understand!

tamaraaPrs (approved):

PR #114, opened 2023-04-28
Origin/refractor

tamaraaPrs:
This PR is used to refractor the different packages so that the classes are better arranged and to define a package structure that we will use until the end.

The project is composed of 4 main packages:
- `base`: Its role is to translate into classes the basic conceptual elements that constitute our game and more generally our application (Player, User, Location, Game... etc). **There is no UI in this package**, but the different elements can interact with each other, as it could interact conceptually, without consideration of the UI.
- `network`: Regroups all that concerns the DB and classes to interact with the DB
- `ui`: Constitutes the entire graphical interface of our application, all activities and composables. _remarks_: In this package there is a `theme` sub-package with all the UI elements common to the whole application like default colors, default padding, etc.
- `utils`: Everything that is useful for our entire application and packages (like gloables instances)


jadtl (approved):
Great file tree! The packages really are neater.

As a nitpick, I'm wondering whether the `rules_and_lobby` package makes sense as they combine two things and why `RulesObject.kt` isn't inside this package. I think there are few enough files so that they can all be at the root of `menu`.

Thanks!

vrooomer (approved):
As we already discussed, this refactor is very good! Thanks for taking your time to do it in a proper fashion.

~ v:)

tamaraaPrs:
> Great file tree! The packages really are neater.
> 
> As a nitpick, I'm wondering whether the `rules_and_lobby` package makes sense as they combine two things and why `RulesObject.kt` isn't inside this package. I think there are few enough files so that they can all be at the root of `menu`.
> 
> Thanks!

I think the problem is mostly the names, not the packages and files disposition 😅
`RulesObject.kt` denotes the rules of the game in general that are displayed in the menu, the rules I'm talking about in the `rules_and_lobby` package are the rules defined by the lobby. But I think that the name `parameter` for the lobby rules might be more appropriate and would make the difference. What do you think of this new name? And you @vrooomer ?

jadtl:
Thanks for clarifying @tamaraaPrs, I agree to rename the game rules to game parameters, that's a good idea, and then maybe the `rules_and_lobby` package could simply be called `lobby`?

tamaraaPrs:
> Thanks for clarifying @tamaraaPrs, I agree to rename the game rules to game parameters, that's a good idea, and then maybe the `rules_and_lobby` package could simply be called `lobby`?

yes, this is what I said in my previous comment

jadtl:
Oh right my bad haha.

vrooomer:
Totally agree! :)

PR #117, opened 2023-04-29
Refactor heads-up display

jadtl:
The `MapActivity` (soon `GameActivity`) is getting pretty cluttered, so I refactored all the composables specific to the HUD into a separate file. 

By the way, while doing that I've noticed that the location passed to `HudLocation` should be a parameter of `Hud` instead of hardcoding it.


AmLibra (approved):
Not much has changed, good to go.

PR #118, opened 2023-04-29
Fix game composables z-order

jadtl:
This is a straight-forward fix for the roll dice button being incorrectly displayed on top of the game menu button.

> In Compose, the order in which you add composables to a parent determines their z-order, with later composables appearing on top of earlier ones.

)


AmLibra (approved):
All good. I'm planning on reworking the UI anyways.

PR #123, opened 2023-04-30
Feature/guest

tamaraaPrs:
Offers the possibility to open the game as a guest, not by joining. 
So in this PR I added :
- A button in `SigniInActivity` to play as a guest.
- An activity named `CreateProfileActivity` to quickly create a profile when the user doesn't have one at all (either because he plays as a guest or because he logs in for the first time)
- The "play as guest" button leads to the `CreateProfileActivity`

### DIFF
--- a/app/src/main/java/com/github/polypoly/app/ui/menu/SignInActivity.kt
+++ b/app/src/main/java/com/github/polypoly/app/ui/menu/SignInActivity.kt
@@ -127,7 +127,7 @@ class SignInActivity : ComponentActivity() {
                     // The first element is the logo of the game
                     GameLogo()
                     Spacer(modifier = Modifier.weight(1f))
-                    SignOrGuestButtons()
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/github/polypoly/app/ui/menu/SignInActivity.kt
```diff
@@ -112,12 +127,15 @@ class SignInActivity : ComponentActivity() {
                     // The first element is the logo of the game
                     GameLogo()
                     Spacer(modifier = Modifier.weight(1f))
-                    SignInButton()
+                    SignOrGuestButtons()
```
AmLibra, line 130: Should probably be SignInOrGuestButtons

AmLibra (approved):
Nothing much to say, quality code and documentation. Extensive testing and looks like it works properly.

PR #124, opened 2023-04-30
Create Constants.kt

Kaz-ookid:
A file where to store constants to avoid magic  numbers and allows easy scaling for our application.


AmLibra (approved):
LGTM!

PR #126, opened 2023-05-01
Feat/create game lobby

Kaz-ookid:
This PR adds an activity to create game lobbies by choosing among game settings!
It is linked to the "Create a game" button, in **WelcomeActivity**.
When pressing the button to create the lobby, it sends to the GameLobbyActivity.

Currently, it does create a lobby but it does not add it to the DB. (reason: waiting for the PR that refactors the DB)
Thus, it sends to the code "1234" lobby for now.

This PR also adds a **UniqueCodeGenerator**,  class that holds functions to create a unique game code.

Tests are not written yet, that is why it is a Draft PR. But reviews about the actual code are welcome

### DIFF
--- a/app/src/main/java/com/github/polypoly/app/base/menu/lobby/GameParameters.kt
+++ b/app/src/main/java/com/github/polypoly/app/base/menu/lobby/GameParameters.kt
@@ -8,7 +8,7 @@ import com.github.polypoly.app.base.game.location.LocationRepository
  * @property gameMode The game mode of the future game
  * @property minimumNumberOfPlayers The minimum number of players needed to start the game
  * @property maximumNumberOfPlayers The maximum number of players that can join the lobby
- * @property roundDuration The duration of a round in the
+ * @property roundDuration The duration of a round in minutes
  * @property maxRound The maximum number of round before the game end. This settings is available
  * only in RICHEST_PLAYER mode. Is null if an other game mode is selected.
  * @property gameMap The map of the game whit the different zones available
--- a/app/src/main/java/com/github/polypoly/app/ui/menu/lobby/CreateGameLobbyActivity.kt
+++ b/app/src/main/java/com/github/polypoly/app/ui/menu/lobby/CreateGameLobbyActivity.kt
@@ -1,0 +1,294 @@
+package com.github.polypoly.app.ui.menu.lobby
+
+import android.os.Bundle
+import androidx.activity.compose.setContent
+import androidx.compose.foundation.layout.*
+import androidx.compose.material.*
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.ArrowForward
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.min
+import com.github.polypoly.app.R
+import com.github.polypoly.app.base.GameMusic
+import com.github.polypoly.app.base.menu.lobby.GameMode
+import com.github.polypoly.app.ui.menu.MenuActivity
+import com.github.polypoly.app.ui.theme.PolypolyTheme
+import com.github.polypoly.app.ui.theme.UIElements
+import com.github.polypoly.app.utils.global.GlobalInstances.Companion.uniqueCodeGenerator
+
+class CreateGameLobbyActivity :  MenuActivity("Create a game") {
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        setContent {
+            MenuContent{
+                CreateGameLobbyContent()
+            }
+        }
+    }
+
+    @Composable
+    fun CreateGameLobbyContent() {
+
+        PolypolyTheme {
+            GameSettingsMenu()
+        }
+    }
+
+    /**
+     *  this function creates a menu that allows the user to chose settings for the game.
+     *  The user can chose the name of the game, if the game is private, the min number of players, the max number of players, the number of rounds, the round duration, the GameMode, the initialPlayerBalance.
+     *  these are displayed as a list of fields that the user can edit, with arrows on the right and on the left to iterate through the values (not for the game name).
+     *  there is also a button at the end to create a lobby, that is disabled until all the fields are filled.
+     *  above the button will be shown the game code, that will be used by other players to join the lobby. The game code should be "FAKE_CODE" for now
+     *
+     *  all the fields title should stick to the left side, all the editable fields should stick to the right side.
+     *  the menu should look like this ('[]' are used to indicate a field that can be edited):
+     *  [Choose a game name] (text field)
+     *  Private game :          [] (checkbox. default is false)
+     *  Min number of players : [] (arrow left) 2 (arrow right) (increases and decreases by 1. min value is 2 and max value is 8. default is 2. must be <= max number of players, otherwise locks value if trying to increase)
+     *  Max number of players : [] (arrow left) 8 (arrow right) (increases and decreases by 1. min value is 2 and max value is 8. Must be >= min number of players. default is 8, otherwise locks the value if trying to decrease)
+     *  Number of rounds :      [] (arrow left) 5 (arrow right) (increases and decreases by 1. min value is 2 and max value is 30. default is 5)
+     *  Round duration :        [] (arrow left) 1 day (arrow right) (5, 10, 15, 20, 25, 30, 45, 60 minutes, then 1, 2, 3, 4, 5, 10, 15, 20 hours, then 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 days. default is 1 day. values will be prompted as so in the field, but will have numerical values in code)
+     *  Game mode :             [] (arrow left) Classic (arrow right) (iterates through GameMode.values(). default GameMode.RICHEST_PLAYER)
+     *  Initial player balance: [] (arrow left) 1500 (arrow right) (increases and decreases by 100. min value is 100 and max value is 10000. default is 1500)
+     *
+     *  Game code is: #FAKE_CODE
+     *  [Create game] (button) (disabled until all the fields are filled)
+     */
+    @Composable
+    fun GameSettingsMenu() {
+        val gameName = remember { mutableStateOf("") }
+        val isPrivateGame = remember { mutableStateOf(false) }
+        val minNumPlayers = remember { mutableStateOf(2) }
+        val maxNumPlayers = remember { mutableStateOf(8) }
+        val numRounds = remember { mutableStateOf(5) }
+        val roundDuration = remember { mutableStateOf("1 day") }
+        val gameMode = remember { mutableStateOf(GameMode.RICHEST_PLAYER) }
+        val initialPlayerBalance = remember { mutableStateOf(1500) }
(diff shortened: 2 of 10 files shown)


app/src/main/java/com/github/polypoly/app/ui/menu/lobby/CreateGameLobbyActivity.kt
```diff
@@ -0,0 +1,334 @@
...
+import com.github.polypoly.app.utils.Constants.Companion.GAME_LOBBY_MENU_PICKER_WIDTH
+import com.github.polypoly.app.utils.Constants.Companion.GAME_LOBBY_MIN_PLAYERS
+import com.github.polypoly.app.utils.Constants.Companion.GAME_LOBBY_MIN_ROUNDS
+import com.github.polypoly.app.utils.Constants.Companion.GAME_LOBBY_PRIVATE_DEFAULT
+import com.github.polypoly.app.utils.Constants.Companion.GAME_LOBBY_ROUNDS_DEFAULT
+import com.github.polypoly.app.utils.Constants.Companion.GAME_LOBBY_ROUNDS_DURATIONS
+import com.github.polypoly.app.utils.Constants.Companion.GAME_LOBBY_ROUND_DURATION_DEFAULT
+import com.github.polypoly.app.utils.global.GlobalInstances.Companion.currentUser
+import com.github.polypoly.app.utils.global.GlobalInstances.Companion.uniqueCodeGenerator
+import com.github.polypoly.app.utils.Constants.Companion.GAME_LOBBY_MIN_INITIAL_BALANCE as GAME_LOBBY_MIN_INITIAL_BALANCE1
```
vrooomer, line 41: I believe there is a mistake there with the importation

app/src/main/java/com/github/polypoly/app/ui/menu/lobby/CreateGameLobbyActivity.kt
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    /**
+     * The menu for creating a game
+     */
+    @Composable
+    fun GameSettingsMenu() {
+        val mContext = LocalContext.current
+
+        val gameName = remember { mutableStateOf("") }
```
vrooomer, line 69: As you only use the `MutableState`s in this function, assigning with `by` instead of `=` would be a better idea, you wouldn't need to call `.value()` each time. If there is another reason for your choice, please tell me :)

app/src/main/java/com/github/polypoly/app/ui/menu/lobby/CreateGameLobbyActivity.kt
```diff
@@ -0,0 +1,334 @@
...
+                contentDescription = "Arrow",
+                modifier = Modifier.size(24.dp).rotate(if (leftArrow) 180f else 0f)
+            )
+        }
+    }
+
+    /**
+     *  Creates a game lobby, registers it in DB and navigates to the game lobby screen
+     */
+    private fun createGame(mContext : Context, name: String, isPrivate: Boolean, minPlayers: Int, maxPlayers: Int, numRounds: Int, roundDuration: String, gameMode: GameMode, initialPlayerBalance: Int, gameCode: String) {
```
vrooomer, line 300: I don't think the name `createGame` is appropriate, you should consider renaming it to `createGameLobby`.

Also, passing all these arguments when you can just pass `GameParameters` is a bit confusing, having too much parameters in a method should be avoided.

app/src/main/java/com/github/polypoly/app/utils/Constants.kt
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Constants used in the app
+ */
+class Constants {
+
+    companion object {
+        /**
+         * Constants for the code generator
+         */
+        const val GAME_LOBBY_CODE_LENGTH = 5
```
vrooomer, line 14: Are all these constants used in a different class than `CreateGameLobbyActivity`? If not, they shouldn't be here.

app/src/main/java/com/github/polypoly/app/utils/UniqueCodeGenerator.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.github.polypoly.app.utils.Constants.Companion.GAME_LOBBY_MAX_CHARACTERS
+import com.github.polypoly.app.utils.global.GlobalInstances
+import com.github.polypoly.app.utils.global.Settings
+import java.util.concurrent.CompletableFuture
+import kotlin.random.Random
+
+/**
+ * Generates a unique code for a game lobby
+ */
+class UniqueCodeGenerator(private val codeLength: Int = GAME_LOBBY_CODE_LENGTH) {
```
vrooomer, line 18: What does the lobby has to do here? I believe you either should rename the class to `UniqueLobbyCodeGenerator` then, however this logic is global and could be very useful if we want to create other unique codes, it is up to you to choose :)

app/src/main/java/com/github/polypoly/app/utils/UniqueCodeGenerator.kt
```diff
@@ -0,0 +1,53 @@
...
+            code = generateCode()
+        } while (!codeIsValid(code))
+        return code
+    }
+
+    /**
+     * Generates a random code of length [codeLength]
+     */
+    private fun generateCode(): String {
+        val digits = GAME_LOBBY_MAX_CHARACTERS
```
vrooomer, line 32: Same here, having `GAME_LOBBY_MAX_CHARACTERS` here seems very unnatural and breaks the abstraction

app/src/main/java/com/github/polypoly/app/utils/UniqueCodeGenerator.kt
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /**
+     * Checks if the code is valid by checking if it is not already in the database
+     */
+    @SuppressLint("NewApi")
+    private fun codeIsValid(code: String): Boolean {
+        var codeIsValid = true
+        val lobbyKey = Settings.DB_GAME_LOBBIES_PATH + code
+        GlobalInstances.remoteDB.keyExists(lobbyKey).thenCompose { keyExists ->
```
vrooomer, line 47: As on a further PR we'll encapsulate all DB operations, please write a TODO on all lines where the DB is explicitly called

vrooomer (changes requested):
The activity looks very cool and easy to use, thanks for your work!

However the code has a lot of abstraction issues that you should fix before pushing them to the main branch.

~ v:)

app/src/main/java/com/github/polypoly/app/ui/menu/lobby/CreateGameLobbyActivity.kt
Kaz-ookid: good idea, I juste had trouble with it so I went for '=' style. Now I found what bothered me and it works well

app/src/main/java/com/github/polypoly/app/utils/Constants.kt
```diff
@@ -0,0 +1,44 @@
...
+        const val GAME_LOBBY_MIN_PLAYERS = 2
+        const val GAME_LOBBY_MAX_ROUNDS = 30
+        const val GAME_LOBBY_MIN_ROUNDS = 2
+        const val GAME_LOBBY_ROUNDS_DEFAULT = 5
+        const val GAME_LOBBY_MIN_INITIAL_BALANCE = 100
+        const val GAME_LOBBY_MAX_INITIAL_BALANCE = 15000
+        const val GAME_LOBBY_INITIAL_BALANCE_STEP = 500
+        const val GAME_LOBBY_INITIAL_BALANCE_DEFAULT = 2500
+        const val GAME_LOBBY_ROUND_DURATION_DEFAULT = "1 day"
+        val GAME_LOBBY_ROUNDS_DURATIONS = mapOf(
```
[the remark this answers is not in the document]
Kaz-ookid: basic knowledge

app/src/main/java/com/github/polypoly/app/utils/Constants.kt
```diff
@@ -0,0 +1,44 @@
...
+        const val GAME_LOBBY_MAX_PLAYERS = 8
+        const val GAME_LOBBY_MIN_PLAYERS = 2
+        const val GAME_LOBBY_MAX_ROUNDS = 30
+        const val GAME_LOBBY_MIN_ROUNDS = 2
+        const val GAME_LOBBY_ROUNDS_DEFAULT = 5
+        const val GAME_LOBBY_MIN_INITIAL_BALANCE = 100
+        const val GAME_LOBBY_MAX_INITIAL_BALANCE = 15000
+        const val GAME_LOBBY_INITIAL_BALANCE_STEP = 500
+        const val GAME_LOBBY_INITIAL_BALANCE_DEFAULT = 2500
+        const val GAME_LOBBY_ROUND_DURATION_DEFAULT = "1 day"
```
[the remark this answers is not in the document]
Kaz-ookid: actually I was about to change it to an enum, it should be way cleaner. I just did this map to quickly work on the UI in the first place, but thanks for reminding, I probably was gonna forget about it haha

app/src/main/java/com/github/polypoly/app/utils/Constants.kt
Kaz-ookid: This one yes (will be used in `JoinGameLobby `to set a max character to code textfield). Anyway for any other constants, they may or may not be used someday elsewhere. But even in the case that they are only used in `CreateGameLobbyActivity`, I think having all important values ("constants") for our code not dispatched everywhere is a good idea, to easely monitor those in one place. I think it is subject to debate. Maybe we can create specific Constants files in packages to have a better abstraction, like GameLobbyConstants or else... For now I will let these here and I suggest we decide something about it!

app/src/main/java/com/github/polypoly/app/utils/UniqueCodeGenerator.kt
Kaz-ookid: I don't see where we would need to generate any other code than for lobby codes. We could imagine making this class modular to generate any type of code, but I don't think it is a necessary work yet. But you are right about the name not being precise, i will change it! :)

app/src/main/java/com/github/polypoly/app/utils/UniqueCodeGenerator.kt
Kaz-ookid: typo, thanks

app/src/main/java/com/github/polypoly/app/base/menu/lobby/GameParameters.kt
```diff
@@ -18,21 +19,18 @@ data class GameParameters (
...
-    // TODO: define real value as this is useless
-    private val maxRoundHours = Int.MAX_VALUE
-
     init {
         if (minimumNumberOfPlayers <= 1)
             throw java.lang.IllegalArgumentException("At least 2 players are needed for a game (provided $minimumNumberOfPlayers)")
         if (maximumNumberOfPlayers < minimumNumberOfPlayers)
             throw java.lang.IllegalArgumentException("Maximum number of players $maximumNumberOfPlayers must be greater than the minimum number $minimumNumberOfPlayers")
-        if (roundDuration <= 0 || roundDuration >= maxRoundHours)
+        if (roundDuration <= 0 || roundDuration >= GameLobbyConstants.RoundDurations.values().last().toMinutes())
```
vrooomer, line 33: Maybe instead of assuming the last value will always be the largest, create a method `RoundDurations.maxRound()` or something like that?

vrooomer (approved):
Really great job on all the changes! I just requested one more change if you want to check it :)

app/src/main/java/com/github/polypoly/app/base/menu/lobby/GameParameters.kt
Kaz-ookid: This enum is supposed to be order so the last will always be the greatest. But I guess a maxRounds can be done to make things clearer

PR #127, opened 2023-05-02
Feature/visit_map

tamaraaPrs:
This PR allows you to visit the map offline, i.e. see the description of each building, see the positive and negative points of the building, etc. All this is `VisitMapActivity`.
This PR also makes the link between `VisitMapActivity` and `GuestMenuActivity`, and refactor the composables and functions around the maps.

### DIFF
--- a/app/src/androidTest/java/com/github/polypoly/app/map/MapViewModelTest.kt
+++ b/app/src/androidTest/java/com/github/polypoly/app/map/MapViewModelTest.kt
@@ -1,7 +1,7 @@
 package com.github.polypoly.app.map
 
 import androidx.arch.core.executor.testing.InstantTaskExecutorRule
-import com.github.polypoly.app.ui.game.GameViewModel
+import com.github.polypoly.app.ui.map.MapViewModel
 import junit.framework.TestCase.assertEquals
 import kotlinx.coroutines.ExperimentalCoroutinesApi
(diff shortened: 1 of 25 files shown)


app/src/androidTest/java/com/github/polypoly/app/map/MapViewModelTest.kt
```diff
@@ -12,7 +12,7 @@ import org.junit.Test
 
 @OptIn(ExperimentalCoroutinesApi::class)
 @InternalCoroutinesApi
-class GameViewModelTest {
+class MapViewModelTest {
```
AmLibra, line 15: Not completely sure to understand why reverting to the old name, I'm assuming because it is now not the case that the ViewModel is dedicated to the GameActivity class ?

AmLibra (approved):
Nothing much to declare. Some architectural changes to the MapActivity class, which is necessary in order to take advantage of the versatility of the MapView component. Code is clean as usual and well documented. Good job and I hope the merging won't be too much of a nightmare.

app/src/androidTest/java/com/github/polypoly/app/map/MapViewModelTest.kt
tamaraaPrs: I changed the name because I use it in a part which is not the game and that the interior of this class concerns only the map in general, there is nothing specific to the game

PR #131, opened 2023-05-03
Encapsulation of DB Objects & Listening 

vrooomer:
This quite big PR is separated in two parts:

### `StorableObject<T>`
This abstract class has one unique goal: to stop explicitly using `RemoteDB` in activities' code.
From now on, every class that represent an object that is going to be stored in the DB should inherit `StorableObject<T>`.

This class provides the classic DB queries that we'll need such as `get()`, `getAll()`, `update()`,... but also more advanced methods such as `onChange()` (everything is documented if you want to check it out).

The reason of the `abstract` class in because each class that inherits `StorableObject<T>` has two "versions":
- The _local_ version: this version is the one that is used in the code, such as our `GameLobby` for example
- The _DB_ version: simple the version that is going to be stored directly in the DB (so a smaller version, only with necessary data in terms of Database systems)

The abstract methods are simply the ones that will define how to convert from `StorableObject<T>` to `T` and vice-versa.

_PS_: I couldn't find any way to make `StorableObject<T>` with abstract static methods so some functions such as `get()` aren't static even if they could..

_PPS_: As most of the methods directly call `RemoteDB` methods without any added logic, I only made tests for `get()` and `getAll()`.

_PPPS_: If you want to know how to correctly use `StorableObject<T>`, I strongly recommend you to check out `StorableObjectTest`.

### `RemoteDB` & `IRemoteStorage`
To create a complete version of `StorableObject<T>`, I had to implement new methods: `removeValue()`, `addChangeListener()`, `removeChangeListener()`. The names are pretty explicit so I won't explain what they do unless it is really necessary (everything is well documented).

_PS_: I added all the corresponding tests in `RemoteDBTest`, however I did not implement the new methods in `MockDB`, I added TODOs

~ v:)


app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
```diff
@@ -105,4 +107,66 @@ open class RemoteDB(
...
+                action(rootRef.child(key))
+            }
+            CompletableFuture.completedFuture(exists)
+        }
+    }
+
+    private val changeListeners = mutableMapOf<String, ValueEventListener>()
+    private val removeListeners = mutableMapOf<String, ValueEventListener>()
+
+    override fun <T : Any> addChangeListener(
```
Jharaxus, line 139: I don't think that such design is very safe on a large application viewpoint: if only one listener can be added per key, what if two actors want to listen a given key at the same time? Even if no such case is explicitly intended, `addChangeListener` still has the huge side effect of clearing the potentially previously stored listener without any ways of knowing it.
This looks quite error prone as it shares a responsibility across actors that do not know each others and could cause very hard to find bugs...
Maybe you could instead support a list of listeners waiting for a change, and add an optional tag when registering a new one so that the user can remove it using the same tag without affecting any other listener still pending (it's a bit more complex but much safer and flexible, and that's how firebase work as you can add any number of listeners to a single DatabaseReference = to a single key).

Jharaxus (changes requested):
Nice work, it looks like it took you a lot of time to make all this components interact in a correct way!
Actually, I am not entirely convinced by the new interface that is exposed here: if I understood correctly, the idea is that the "storage" concept is hidden behind the classes inheriting `StorableObject`, and you can call `Object().get(key)` to have a future that holds the result from remoteDB.
This indeed has the benefit of simplifying the usage of `RemoteDB`, especially avoiding to add the root path of the object every time, but I believe that it also has a huge pitfall: it breaks the separation between the data stored and the storage itself.
Here, you have no choice but to use the storable objects with the remoteDB, but what if you also want to store them in another storage (such as a local device storage for persistency across app usages for example)? It would be heavy to create a StorableObjectX for each type of storage that will use it...
This coupling too strong between these two concepts is actually revealed by some elements in this code: you can't put all the `get`, `getAll`, etc.. functions static as you would like, and they effectively are mostly wrappers around the storage behind which is a remoteDB. Another way to see that I see a friction is with calls such as `TestPerson().get(LOCAL_KEY)`: where is the data it fetch at during the get? Is it a costly operation or not? I don't think that the gain in usability compensates the huge loss in meaning (even more so as the functions cannot be static).
Sorry, that may have been a bit harsh but I tried as best as I could to express why I don't find this architecture entirely satisfying, even if you clearly worked a lot to make it work.
Now here is a suggestion regarding how to adapt a bit this framework to solve the above issues while avoiding having to change too many things:
The idea would be to clearly differentiate data, which knows it can be stored, and the actual storage:
- StorableObject could be simplified and only keep the functions `toDBObject()` and `toLocalObject(dbObject: T)` as well as the path attribute --> it knows it can be stored and provides helper functions to do so, but it doesn't care about the storing logic (kind of like the `serialize`/`deserialize` interface that all objects implement in java). It would still be inherited by another object wishing to be stored in a storage (GameLobby, Game, etc...)
- IRemoteStorage (or remoteDB for the implementation) can keep its current interface but would add a constraint to the generic type which would have to inherit StorableObject (for example: `override fun <T, U: StorableObject<T>> updateValue(key: String, value: U) ...`). It would give access to the above mentioned storage object's helper functions in the storage's implementation, but each storage would be able to choose how to use it (clear separation between data and storage).
- Possibly later on, another abstract function `getKey()` could be added to `StorableObject` for the storage to use. This would let it offer an interface where the user can even avoid to explicitly specify the key.

Finally the end, that was a huge comment 😅
I would first be happy to know what you think about my analysis of the current interface, the rest is only a proposition for if we reach the consensus that things could be improved :)

vrooomer:
Thanks for the huge review!
Actually, I've been struggling to find the best way to encapsulate the DB queries and your comment is like very interesting.

I never thought of directly modifying `IRemoteStorage` and actually it could be a very good idea. I think I understood your points and I have to agree that it seems like a decent solution to these problems.

As the new changes are pretty big (would involve modify tests, interfaces, etc.) and as we're at the end of the sprint, I'll implement it in the next sprint. It shouldn't take that much time as all the logic is already implemented!

I have a lot of new ideas that match that solution and I'll let you know in the next sprint 😃

~ v:)

app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
vrooomer: Good point! I'll add a tag system so that we can have multiple listeners for a given `DataReference`

app/src/androidTest/java/com/github/polypoly/app/network/RemoteDBTest.kt
```diff
@@ -204,4 +205,183 @@ class RemoteDBTest: PolyPolyTest(false, false) {
...
+        val data2 = TEST_USER_2
+        addDataToDB(data1, key)
+
+        remoteDB.addChangeListener(key, action1, User::class).get(TIMEOUT_DURATION, TimeUnit.SECONDS)
+        remoteDB.addChangeListener(key, action2, User::class).get(TIMEOUT_DURATION, TimeUnit.SECONDS)
+        remoteDB.updateValue(key, data2).get(TIMEOUT_DURATION, TimeUnit.SECONDS)
+        remoteDB.updateValue(key, data1).get(TIMEOUT_DURATION, TimeUnit.SECONDS)
+        remoteDB.updateValue(key, data2).get(TIMEOUT_DURATION, TimeUnit.SECONDS)
+        assertTrue(num2 >= 3)
+    }
```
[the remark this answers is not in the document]
vrooomer: No actually purpose of this test is to check if the **new** listener is called after the overwrite.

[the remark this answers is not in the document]
Jharaxus: Oh you're right I didn't get the test name right 😅

vrooomer:
I changed the little issues you pointed, the big task will be done next sprint :)

vrooomer:
As we need to discuss on the DB query structure, we won't merge this PR (even if it's ready) during this sprint, so turned it into a _Draft PR_ so there is no ambiguity.

vrooomer:
_As the PR was initially open and then modified, I'll copy and paste the initial description with a few changes_

This quite big PR is separated in two parts:

### `StorableObject<T>`
This abstract class has one unique goal: to restrict all DB objects to respect some rules, so that in a further PR, `RemoteDB` will only allow operations on subclasses of `StorableObject<T>`.

Each time a subclass is created, the companion object of `StorableObject` will store the conversion method and the DB path associated to this new class. For implementation details, check the code and let me know if something isn't clear!

The reason of the `abstract` class in because each class that inherits `StorableObject<T>` has two "versions":
- The _local_ version: this version is the one that is used in the code, such as our `GameLobby` for example
- The _DB_ version: simple the version that is going to be stored directly in the DB (so a smaller version, only with necessary data in terms of Database systems)

The abstract methods are simply the ones that will define how to convert from `StorableObject<T>` to `T` and vice-versa.

_PS_: If you want to know how to correctly use `StorableObject<T>`, I strongly recommend you to check out `StorableObjectTest`.

### `RemoteDB` & `IRemoteStorage`
To create a complete version of `StorableObject<T>`, I had to implement new methods: `removeValue()`, `addOnChangeListener()`, `removeOnChangeListener()`. The names are pretty explicit so I won't explain what they do unless it is really necessary (everything is well documented).

_PS_: I added all the corresponding tests in `RemoteDBTest`, however I did not implement the new methods in `MockDB`, I added TODOs

~ v:)

app/src/androidTest/java/com/github/polypoly/app/commons/MockDB.kt
```diff
@@ -19,9 +20,9 @@ class MockDB: IRemoteStorage {
     override fun <T : Any> getValue(key: String, clazz: KClass<T>): CompletableFuture<T> {
         val keyCleaned = cleanKey(key)
         if (!data.containsKey(keyCleaned)) {
-            val failure = CompletableFuture<T>()
-            failure.completeExceptionally(IllegalAccessException("Invalid key $keyCleaned"))
-            return failure
+            val failedPromise = CompletableFuture<T>()
```
SolalPirelli, line 23: I get that promises and futures are kind of the same thing but since you're using a class explicitly named Future, use names that match?

app/src/androidTest/java/com/github/polypoly/app/commons/MockDB.kt
```diff
@@ -19,9 +20,9 @@ class MockDB: IRemoteStorage {
...
+            failedPromise.completeExceptionally(IllegalAccessException("Invalid key $keyCleaned"))
```
SolalPirelli, line 24: `NoSuchElementException`? IllegalAccess is for reflection (I realize this is existing code)

app/src/androidTest/java/com/github/polypoly/app/commons/MockDB.kt
```diff
@@ -71,6 +73,28 @@ class MockDB: IRemoteStorage {
         return CompletableFuture.completedFuture(true)
     }
 
+    override fun removeValue(key: String): CompletableFuture<Boolean> {
+        TODO("Not yet implemented")
+    }
+
+    override fun <T : Any> addOnChangeListener(
+        key: String,
+        tag: String,
```
SolalPirelli, line 82: What's the point of the tag? It doesn't seem used in this PR (maybe I missed something)

app/src/androidTest/java/com/github/polypoly/app/commons/PolyPolyTest.kt
```diff
@@ -24,6 +25,7 @@ import kotlinx.coroutines.launch
 import org.junit.After
 import org.junit.Before
 import org.junit.runner.RunWith
+import org.mockito.Mockito.`when`
```
SolalPirelli, line 28: Unused imports? Since there are no changes in this file that use them

app/src/androidTest/java/com/github/polypoly/app/network/RemoteDBTest.kt
```diff
@@ -16,27 +17,31 @@ import java.util.*
...
+    private val testDB: RemoteDB
 
     init {
         val db = Firebase.database
         try {
             db.setPersistenceEnabled(false)
         } catch (_: Exception) {}
-        remoteDB = RemoteDB(db, rootTests)
+        testDB = RemoteDB(db, rootTests)
+        remoteDB = testDB
```
SolalPirelli, line 36: why have both remoteDB and testDB fields?

app/src/androidTest/java/com/github/polypoly/app/network/RemoteDBTest.kt
```diff
@@ -204,4 +209,268 @@ class RemoteDBTest: PolyPolyTest(false, false) {
...
+        val action = { _: User -> num += 1 }
+
+        val data1 = TEST_USER_1
+        val data2 = TEST_USER_2
+        addDataToDB(data1, key)
+
+        remoteDB.addOnChangeListener(key, "tag", action, User::class).get(TIMEOUT_DURATION, TimeUnit.SECONDS)
+        remoteDB.updateValue(key, data2).get(TIMEOUT_DURATION, TimeUnit.SECONDS)
+        remoteDB.updateValue(key, data1).get(TIMEOUT_DURATION, TimeUnit.SECONDS)
+        assertTrue(num >= 2)
```
SolalPirelli, line 283: check that the callback is executed with the right data? you could replace the counter by a list and use Hamcrest's `contains` matcher

app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
```diff
@@ -15,7 +17,7 @@ open class RemoteDB(
     root: String
 ) : IRemoteStorage {
```
SolalPirelli, line 18: why is this class `open`?

app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
```diff
@@ -105,4 +107,77 @@ open class RemoteDB(
...
+                ref.removeEventListener(previousListener)
+            }
+        }
+    }
+
+    override fun deleteAllOnChangeListeners(key: String): CompletableFuture<Boolean> {
+        return getRefAndThen(key) { ref ->
+            changeListeners
+                .filterKeys { pair -> pair.first == key }
+                .map { (pair, _) -> pair }
```
SolalPirelli, line 173: what is this doing?

app/src/main/java/com/github/polypoly/app/network/storable/StorableObject.kt
```diff
@@ -0,0 +1,104 @@
...
+ * It is the role of every subclass to define the path and the conversion methods.
+ *
+ * This class takes a generic type T as argument, corresponding to the data class that is
+ * going to be stored in the DB.
+ *
+ * @note attribute [key] should be the primary key used to store the object in the DB
+ *
+ * @note for the subclass to be correctly initialized, one instance should be created before any operation
+ */
+abstract class StorableObject<T : Any> (dbPath: String, private val key: String) {
```
SolalPirelli, line 21: should `dbPath` be private as well?

app/src/main/java/com/github/polypoly/app/network/storable/StorableObject.kt
```diff
@@ -0,0 +1,104 @@
...
+ * - The converter (to DB)]that converts an instance of a local StorableObject<T> to a DB T.
+ *
+ * It is the role of every subclass to define the path and the conversion methods.
+ *
+ * This class takes a generic type T as argument, corresponding to the data class that is
+ * going to be stored in the DB.
+ *
+ * @note attribute [key] should be the primary key used to store the object in the DB
+ *
+ * @note for the subclass to be correctly initialized, one instance should be created before any operation
```
SolalPirelli, line 19: what does this mean? how could you use the subclass without creating an instance of it?

app/src/main/java/com/github/polypoly/app/network/storable/StorableObject.kt
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * [StorableObject] is the super class of all classes that'll be contained in the DB.
+ * It implements all the abstractions needed by the remote storage to do all the queries, i.e.
+ * - The [dbPath] that precedes every key for a given class (e.g. "objectX/"),
+ * - The converter (to local) that converts a T DB object to a local StorableObject<T>,
+ * - The converter (to DB)]that converts an instance of a local StorableObject<T> to a DB T.
+ *
+ * It is the role of every subclass to define the path and the conversion methods.
+ *
+ * This class takes a generic type T as argument, corresponding to the data class that is
```
SolalPirelli, line 14: this is intended to be `class X : StorableObject<X>`, right?

SolalPirelli: ah, no, from the conversions below it's not... so what kind of class is this expected to be used with? should `StorableObject` be an interface instead?

SolalPirelli:
Makes sense overall, a lot of the complexity is a necessary workaround for the lack of abstract statics in Kotlin I guess.

Though it'd be clearer to have at least one real example of a storable object in the PR, which would also serve as a sanity check that it can actually be used for the purposes it intends to be used.

app/src/main/java/com/github/polypoly/app/network/storable/StorableObject.kt
AlSchlo: About this, shouldn't the key be the path as well?

AlSchlo: Or it would make more sense to make the path private but the key public?

app/src/androidTest/java/com/github/polypoly/app/network/RemoteDBTest.kt
```diff
@@ -204,4 +209,268 @@ class RemoteDBTest: PolyPolyTest(false, false) {
...
+    @Test
+    fun registeredDataIsRemoved() {
+        val key = "some_key"
+        val data = TEST_USER_1
+        remoteDB.registerValue(key, data).get(TIMEOUT_DURATION, TimeUnit.SECONDS)
+        remoteDB.removeValue(key).get(TIMEOUT_DURATION, TimeUnit.SECONDS)
+        assertFalse(
+            remoteDB.keyExists(key).get(TIMEOUT_DURATION, TimeUnit.SECONDS)
+        )
+    }
```
Jharaxus, line 241: Those two tests could be merged into one as they are very close and test small highly related and interdependent behaviors

app/src/main/java/com/github/polypoly/app/network/storable/StorableObject.kt
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * @note attribute [key] should be the primary key used to store the object in the DB
+ *
+ * @note for the subclass to be correctly initialized, one instance should be created before any operation
+ */
+abstract class StorableObject<T : Any> (dbPath: String, private val key: String) {
+
+    init {
+        registerClassToCompanion(dbPath, this::class.toString(), ::toLocalObject)
+    }
```
Jharaxus, line 25: This should be in the init of the companion object instead: here it would be called at each instantiation of a `StorableObject`

Jharaxus (changes requested):
Two little additional comments, otherwise it looks good for the basis of the final DB interface.

app/src/androidTest/java/com/github/polypoly/app/network/RemoteDBTest.kt
vrooomer: The first checks that `removeValue `returns correctly true, the second one that it is indeed removed, even if that means the same thing, two different things are checked and thus having two tests is a better approach.

app/src/main/java/com/github/polypoly/app/network/storable/StorableObject.kt
vrooomer: This should be called at the initialization of each of the subclasses, having that block in the companion object of `StorableObject` would call it once and thus would force the subclasses to have a companion object, we want to avoid this.

The block will be called at each instanciation of a subclass but this will only create low overhead and the same thing will be written each time so there is only a very low performance impact.

However if you have another solution, let me know :)

app/src/androidTest/java/com/github/polypoly/app/commons/MockDB.kt
vrooomer: Thanks, it is renamed now

app/src/androidTest/java/com/github/polypoly/app/commons/MockDB.kt
vrooomer: So far it is only used in tests, however it is pretty useful as some DB elements may be listened by different activities for example.

app/src/androidTest/java/com/github/polypoly/app/network/RemoteDBTest.kt
vrooomer: Maybe the names aren't the best. `remoteDB` is used in all the program, whereas `testDB` is instantiated locally. I made this to avoid test conflicts, before, some local DB tests (that use `MockDB`) put data in our real DB so with this assignation, I make sure that before each local/online DB test, the correct value is assigned to `remoteDB`

app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
vrooomer: No idea, this is @Jharaxus code, however i'll remove it, if we need it we'll change it again.

app/src/main/java/com/github/polypoly/app/network/storable/StorableObject.kt
vrooomer: As mentioned in the PR, you can check an example in `StorableObjectTest`, the `T` class in the one that is going to be stored in the DB and the `StorableObject<T>` the local one.

The reason I choose an abstract class was the use of our companion object :)

app/src/main/java/com/github/polypoly/app/network/storable/StorableObject.kt
```diff
@@ -0,0 +1,104 @@
...
+     * Converts the given DB version to its local version
+     * @param dbObject: the object to convert
+     * @return the local version
+     *
+     * @note /!\ STATIC /!\ The returned value shouldn't depend on the current object instance
+     */
+    protected abstract fun toLocalObject(dbObject: T): StorableObject<T>
+
+    // ================================================================== HELPERS
+    protected fun getKey(): String { return key }
```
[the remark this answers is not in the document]
vrooomer: This was to force immutability, however as Java/Kotlin strings are immutable, I'll change it and make `key` public

app/src/main/java/com/github/polypoly/app/network/storable/StorableObject.kt
vrooomer: `dbPath` has no `val` associated to it so no one can access `dbPath`, so it is already a way of private.
The `key` is however meant to be public and was private for immutability reasons as mentioned below but now it is fixed :)

app/src/main/java/com/github/polypoly/app/network/storable/StorableObject.kt
Jharaxus: This is indeed an acceptable trade-off !

app/src/androidTest/java/com/github/polypoly/app/network/RemoteDBTest.kt
vrooomer: I now added a test, thanks :)

PR #134, opened 2023-05-04
Adds spin loader mechanic

Jharaxus:
- Adds a circular loader component to use in UI when loading something
- Adds a LoadingModel ViewModel to abstract the concept of loading data in a ViewModel. It is partly intended to be used with the above-mentioned loader


Kaz-ookid (approved):
Neat addition and looks well implemented, nothing to declare :)

PR #136, opened 2023-05-04
Basic game mechanics in GameActivity

Jharaxus:
Add basic game mechanic with a button that can increase the game round until the game is completed.


app/src/main/java/com/github/polypoly/app/base/game/Game.kt
```diff
@@ -31,13 +31,14 @@ class Game private constructor(
...
         ) } }
-    val currentRound: Int = 1
+    var currentRound: Int = 1
 
     /**
      * Go to the next turn
      */
     fun nextTurn() {
         // TODO update the data with the DB
+        ++currentRound
```
Kaz-ookid, line 41: Maybe we should keep a track of the round number in the db itself, to avoid any player having false round number in local?

app/src/main/java/com/github/polypoly/app/data/GameRepository.kt
```diff
@@ -1,15 +1,28 @@
...
          */
         var gameCode: String? = null
+
+        /**
+         * Game used accross the application
+         */
+        var game: Game? = null
+
+        /**
+         * Game used accross the application
```
Kaz-ookid, line 24: I think you forgot to refactor this comment

app/src/main/java/com/github/polypoly/app/ui/game/GameActivity.kt
```diff
@@ -25,79 +34,84 @@ import org.osmdroid.views.overlay.Marker
...
+    }
+
+    @Composable
+    fun GameActivityContent() {
+        val player = gameModel.getPlayerData().observeAsState().value
+        val game = gameModel.getGameData().observeAsState().value
+        val gameTurn = gameModel.getRoundTurnData().observeAsState().value
+        val gameEnded = gameModel.getGameFinishedData().observeAsState().value
+
+        if (player != null && game != null && gameTurn != null && gameEnded != null) {
```
Kaz-ookid, line 51: What does this iteration of conditions define? Is it to check if a game is occurring ? Maybe for clarity, creating a method with an adequate name would provide easier understanding of the code. But I admit this is neat picking ;)

app/src/main/java/com/github/polypoly/app/ui/game/GameActivity.kt
```diff
@@ -25,79 +34,84 @@ import org.osmdroid.views.overlay.Marker
...
 
+    @Composable
+    fun nextTurnButton(gameEnded: Boolean) {
+        Box(modifier = Modifier.fillMaxWidth()) {
+            Button(
+                modifier = Modifier
+                    .size(30.dp)
+                    .align(Alignment.BottomCenter)
+                    .offset(y = (-30).dp)
+                    .testTag("nextTurnButton"),
```
Kaz-ookid, line 83: Please use snake_case for test tags :)!

app/src/main/java/com/github/polypoly/app/ui/menu/lobby/GameLobbyActivity.kt
```diff
@@ -210,6 +210,15 @@ class GameLobbyActivity : ComponentActivity() {
         }
     }
 
+    private fun launchGameActivity(packageContext: Context) {
+        val completedLobby = gameLobbyWaitingModel.getGameLobby().value
+        if (completedLobby != null) {
+            GameRepository.game = Game.launchFromPendingGame(completedLobby)
+            val gameIntent = Intent(packageContext, GameActivity::class.java)
+            startActivity(gameIntent)
```
Kaz-ookid, line 218: You probably should call finish() to ensure nobody can return to a lobby once a game is started

Kaz-ookid (commented):
Nice additions, code looks very good! Well done :) I’ll Let you finish dealing with tests not passing, good luck!

app/src/main/java/com/github/polypoly/app/ui/game/GameActivity.kt
Jharaxus: Not exactly, these checks solve the fact that the value returned by `observeAsState().value` may be null. This is actually the recommended way to do it safely that I found in the official android doc.

app/src/main/java/com/github/polypoly/app/base/game/Game.kt
Jharaxus: This is an intended behavior, this data should indeed be synchronised across all the players (as the comment above tried to suggest). Here this pr doesn't aim at doing that, but the goal was instead to lay a solid fondation for the data logic/UI interactions that will be needed in the next sprints. I'll update this pr's doc to reflect this intent better :)

Kaz-ookid (approved):

PR #150, opened 2023-05-06
Add player states

AmLibra:
Small edit allowing Hugo and myself to work on a common basis regarding linking the Game and the Database, as well as displaying the appropriate UI when required.


vrooomer (approved):
Exactly corresponds to what we were talking about so LGTM!

~ v:)

PR #151, opened 2023-05-08
Update google-services.json

Kaz-ookid:
updating google-services.json to accept @SolalPirelli and @AlSchlo devices authentication wise


Jharaxus (approved):
This should be fine :)

PR #155, opened 2023-05-15
Use of DB refactored

vrooomer:
After the implementation of `StorableObject` in #131, this class is now used in `IRemoteStorage`.

All the data stored in the DB must now inherit from `StorableObject`, thus all code that use the DB has been refactored and adapted to this change, although features like polling are kept the same as I just refactored and not changed the logic.

I had to navigate through all the code to modify it so this PR is huge but the refactoring isn't very complex.

I also added and modified tests for various classes, especially `StorableObject` and `RemoteDB`

Let me know if you have any issue for the PR!

_Note: we still don't have the possibility to listen to a whole group of objects (e.g. listen to any change in lobbies/), as we only can listen to a specific node. This will be part of a further PR._

~ v:)


app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
```diff
@@ -53,123 +47,191 @@ class RemoteDB(
...
-        return getValueAndThen(key, { data, valuesPromise ->
-            val users = ArrayList<T>()
-            for (child in data.children) {
-                users.add(child.getValue(clazz.java)!!)
+        rootRef.child(absoluteKey).get().addOnSuccessListener { data ->
+            if(!data.exists()) {
+                waitingFuture.complete(onFailure)
+            } else {
+                waitingFuture.complete(onSuccess(data))
             }
```
AmLibra, line 67: maybe something like this ? waitingFuture.complete(if (data.exists()) onSuccess(data) else onFailure)

app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
```diff
@@ -53,123 +47,191 @@ class RemoteDB(
...
-            valuesPromise.complete(users)
-        }, { promise -> promise.complete(listOf()) })
+        }.addOnFailureListener(waitingFuture::completeExceptionally)
+
+        return waitingFuture.thenCompose { future -> future }
```
AmLibra, line 70: I usually like this, less redundancy
return waitingFuture.thenCompose { it }

app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
```diff
@@ -53,123 +47,191 @@ class RemoteDB(
...
+     */
+    private fun getRefAndThen(absoluteKey: String, action: (DatabaseReference) -> Unit): CompletableFuture<Boolean> {
+        return getDataAndThen(
+            absoluteKey,
+            { data ->
+                action(data.ref)
+                CompletableFuture.completedFuture(true)
+            },
+            CompletableFuture.completedFuture(false)
+        )
```
AmLibra, line 96: Have seen this above and below, but you can add the import:
import java.util.concurrent.CompletableFuture.completedFuture
in order to avid repeating CompletableFuture (Intellij will help you with the yellow marks on the sidebar)

app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
```diff
@@ -53,123 +47,191 @@ class RemoteDB(
...
     }
 
-    override fun keyExists(key: String): CompletableFuture<Boolean> {
-        return getValueAndThen(
-            key,
-            { _, promise -> promise.complete(true) }, // value found so key exists
-            { promise -> promise.complete(false) }  // value not found so key doesn't exist
-        )
+    override fun <T : StorableObject<*>> keyExists(key: String, clazz: KClass<T>): CompletableFuture<Boolean> {
+        return absoluteKeyExists(StorableObject.getPath(clazz) + key)
```
AmLibra, line 152: Same import thing for StorableObject

app/src/main/java/com/github/polypoly/app/ui/menu/lobby/UniqueGameLobbyCodeGenerator.kt
```diff
@@ -13,12 +13,15 @@ import kotlin.random.Random
...
-        return code
+    fun generateUniqueGameLobbyCode(): CompletableFuture<String> {
+        val code = generateCode()
+        return codeIsUsed(code).thenCompose { isUsed ->
+            if(isUsed) {
+                generateUniqueGameLobbyCode()
+            } else {
+                CompletableFuture.completedFuture(code)
+            }
+        }
```
AmLibra, line 24: Nitpick, but usually the way this is handled is using something like 'previousUID++', to avoid relying on randomness and ensuring quick and guaranteed result (obviously limited by MAX_INT)

AmLibra (approved):
Overall good code and extensive testing, this code looks like something pleasant to work with. Thank you for your work. Feel free to work on the nitpicks before merging, but I will approve this for now.

app/src/main/java/com/github/polypoly/app/ui/menu/lobby/UniqueGameLobbyCodeGenerator.kt
vrooomer: As this is @Kaz-ookid 's work, I'll leave it like this but I strongly agree with you

app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
vrooomer: Oh yes of course! :)

app/src/main/java/com/github/polypoly/app/ui/menu/lobby/UniqueGameLobbyCodeGenerator.kt
Kaz-ookid: I agree, thanks for the advice I did not think about it. However this works fine for now, as the code is way bigger than needed seeing how many of those codes we will have to produce. This will be to be though about again if we get thousands of players daily i guess haha

app/src/main/java/com/github/polypoly/app/base/menu/lobby/GameLobby.kt
```diff
@@ -18,7 +23,7 @@ data class GameLobby(
     val name: String = "defaultName",
     val code: String = "defaultCode",
     val private: Boolean = false,
```
SolalPirelli, line 25: Can these default values go away now that this class isn't itself stored in the DB?

app/src/main/java/com/github/polypoly/app/base/user/User.kt
```diff
@@ -35,4 +38,13 @@ data class User(
     fun hasTrophy(trophyId: Int): Boolean {
         return trophiesWon.contains(trophyId)
     }
+
+    // ====================================================================== STORABLE
+    override fun toDBObject(): User {
+        return User(id, name, bio, skin, stats, trophiesWon, trophiesDisplay, currentUser)
```
SolalPirelli, line 44: `return this`?

app/src/main/java/com/github/polypoly/app/network/StorableObject.kt
```diff
@@ -95,7 +131,7 @@ abstract class StorableObject<T : Any> (dbPath: String, val key: String) {
      *
      * @note /!\ STATIC /!\ The returned value shouldn't depend on the current object instance
      */
-    protected abstract fun toLocalObject(dbObject: T): StorableObject<T>
+    protected abstract fun toLocalObject(dbObject: T): CompletableFuture<StorableObject<T>>
```
SolalPirelli, line 134: good one-line example of why having the infrastructure + the code that uses it in a single PR can be useful :-)

app/src/main/java/com/github/polypoly/app/ui/menu/rankings/RankingsViewModel.kt
```diff
@@ -5,7 +5,6 @@ import androidx.lifecycle.ViewModel
 import com.github.polypoly.app.base.user.User
 import com.github.polypoly.app.network.getAllValues
 import com.github.polypoly.app.utils.global.GlobalInstances.Companion.remoteDB
-import com.github.polypoly.app.utils.global.Settings.Companion.DB_USERS_PROFILES_PATH
```
SolalPirelli, line 8: isn't it satisfying to see so many times this line removed :D

app/src/main/java/com/github/polypoly/app/network/RemoteDB.kt
SolalPirelli: What is even the point of `thenCompose` here if it's applying the identity function?

PR #160, opened 2023-05-18
Linking user to the DB

vrooomer:
The current user is now finally linked to the DB!
In `GlobalInstances` there is `currentUser` that is available, for simplicity reasons, its key is the uid of the `FirebaseUser` corresponding to the log in information.

I had to modify lot of tests for this new feature to work, but nothing very difficult.

Also, now the id of a `User` is a `String` instead of `Long`

Let me know if you have any question!

~ v:)

### DIFF
--- a/app/src/main/java/com/github/polypoly/app/ui/menu/SignInActivity.kt
+++ b/app/src/main/java/com/github/polypoly/app/ui/menu/SignInActivity.kt
@@ -2,6 +2,7 @@ package com.github.polypoly.app.ui.menu
 
 import android.content.Intent
 import android.os.Bundle
+import android.util.Log
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.compose.foundation.Image
@@ -25,6 +26,7 @@ import com.github.polypoly.app.base.user.User
 import com.github.polypoly.app.utils.global.GlobalInstances.Companion.isSignedIn
 import com.github.polypoly.app.utils.global.GlobalInstances.Companion.remoteDB
 import com.github.polypoly.app.network.StorableObject
+import com.github.polypoly.app.network.addOnChangeListener
 import com.github.polypoly.app.ui.menu.profile.CreateProfileActivity
 import com.github.polypoly.app.ui.theme.PolypolyTheme
 import com.github.polypoly.app.ui.theme.UIElements.MainActionButton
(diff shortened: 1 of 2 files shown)


AmLibra (approved):
It's funny to see how many files were affected by such a little change.
Nothing to declare, approved

app/src/main/java/com/github/polypoly/app/utils/global/GlobalInstances.kt
```diff
@@ -34,12 +36,22 @@ class GlobalInstances {
...
+        var currentUser : User? = null
         var isSignedIn = false
 
+        fun initCurrentUser(key: String) {
+            remoteDB.keyExists<User>(key).thenAccept { exists ->
+                if(exists) {
+                    remoteDB.getValue<User>(key).thenAccept { currentUser = it }
+                } else {
+                    currentUser = User(id = key)
+                    remoteDB.registerValue(currentUser!!)
```
Kaz-ookid, line 49: I suggest to also initialize a simple profile for new users, using their names (that we can get from FBAuthUser). Or get them a random name, but anything to enable differenciation for players if they never change their name! :)

app/src/main/java/com/github/polypoly/app/ui/menu/SignInActivity.kt
```diff
@@ -198,6 +205,31 @@ class SignInActivity : ComponentActivity() {
...
+
+
+
+
+
+
+
+
+
+
```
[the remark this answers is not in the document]
vrooomer: You got me!!!
There is fake data logic there, I don't want to remove it as I didn't create it but it was a bit annoying to have it right next to valid code

Kaz-ookid (approved):
Nice work! I suggested a change but I will approve so that you can decide wether you wish to follow my suggestion or not :)

app/src/main/java/com/github/polypoly/app/utils/global/GlobalInstances.kt
vrooomer: Didn't know about the FBAuthUser name! I'll add it :)

app/src/main/java/com/github/polypoly/app/ui/menu/SignInActivity.kt
```diff
@@ -40,31 +41,47 @@ import java.util.concurrent.CompletableFuture
...
+            initCurrentUser(user.uid, user.displayName ?: "default")
+
+            val welcomeActivityIntent = Intent(this, WelcomeActivity::class.java)
+            isSignedIn = true
+            startActivity(welcomeActivityIntent)
+            finish()
+        }
     }
 
+    @Deprecated("Prefer to use launchWelcomeIfReady")
```
SolalPirelli, line 84: Why deprecate a private function? You could remove it, there's no issue of compatibility here

app/src/main/java/com/github/polypoly/app/ui/menu/SignInActivity.kt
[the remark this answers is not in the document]
SolalPirelli: we do not just give you a grade based on number of lines of code you know :p

PR #161, opened 2023-05-18
Fix max round duration

jadtl:
Just a very simple fix for a bug I encountered while debugging something else: the app would throw an `IllegalArgumentException` if the maximum round duration was set on game lobby creation.


AmLibra (approved):
LGTM

PR #162, opened 2023-05-18
Synchronize players and improve game loop

Jharaxus:
Works on the game loop and gameActivity logic:
- Refactor GameActivity business logic into GameViewModel as much as possible
- Remove dependency to UI-related data in MapViewModel by removing the reference to Markers
- Add responsive synchronization between players (currently only needed at the end of turn)
- Add dice roll and betting in the game loop

### DIFF
--- a/app/src/androidTest/java/com/github/polypoly/app/commons/PolyPolyTest.kt
+++ b/app/src/androidTest/java/com/github/polypoly/app/commons/PolyPolyTest.kt
@@ -3,7 +3,10 @@ package com.github.polypoly.app.commons
 import androidx.lifecycle.LiveData
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.github.polypoly.app.base.game.Player
+import com.github.polypoly.app.base.game.location.LocationProperty
 import com.github.polypoly.app.base.game.location.LocationPropertyRepository
+import com.github.polypoly.app.base.game.location.LocationPropertyRepository.getZones
+import com.github.polypoly.app.base.game.location.Zone
 import com.github.polypoly.app.base.menu.lobby.GameLobby
 import com.github.polypoly.app.base.menu.lobby.GameMode
 import com.github.polypoly.app.base.menu.lobby.GameParameters
@@ -26,6 +29,8 @@ import org.junit.runner.RunWith
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.locks.ReentrantLock
+import kotlin.math.absoluteValue
+import kotlin.random.Random
 
 @RunWith(AndroidJUnit4::class)
 abstract class PolyPolyTest(
@@ -194,5 +199,19 @@ abstract class PolyPolyTest(
         scope.launch { liveData.removeObserver(observer) }
 
         return result
+    }
+
+    fun getRandomLocation(amongLocations: List<LocationProperty> = getZones().flatMap { zone -> zone.locationProperties }): LocationProperty {
+        return amongLocations[Random.nextInt().absoluteValue % amongLocations.size]
+    }
+
+    fun execInMainThread(action: () -> Unit): CompletableFuture<Boolean> {
+        val scope = CoroutineScope(Dispatchers.Main + Job())
+        val future = CompletableFuture<Boolean>()
+        scope.launch {
+            action()
+            future.complete(true)
+        }
+        return future
     }
 }
--- a/app/src/androidTest/java/com/github/polypoly/app/commons/PolyPolyTest.kt
+++ b/app/src/androidTest/java/com/github/polypoly/app/commons/PolyPolyTest.kt
@@ -70,27 +70,27 @@ abstract class PolyPolyTest(
 
         val TEST_GAME_LOBBY_FULL = GameLobby(
             TEST_USER_0, GameParameters(GameMode.RICHEST_PLAYER, 2, 6,
-            60, 20, emptyList(), 100), "Full gameLobby", "lobby1234"
+            60, 20, getZones(), 100), "Full gameLobby", "lobby1234"
         )
         val TEST_GAME_LOBBY_PRIVATE = GameLobby(
             TEST_USER_1, GameParameters(GameMode.RICHEST_PLAYER, 4, 6,
-            360, 20, emptyList(), 300), "Private gameLobby", "lobbyabc123", true
+            360, 20, getZones(), 300), "Private gameLobby", "lobbyabc123", true
         )
         val TEST_GAME_LOBBY_AVAILABLE_1 = GameLobby(
             TEST_USER_1, GameParameters(GameMode.LAST_STANDING, 3, 8,
-            600, null, emptyList(), 1000), "Joinable 1", "lobbyabcd"
+            600, null, getZones(), 1000), "Joinable 1", "lobbyabcd"
         )
         val TEST_GAME_LOBBY_AVAILABLE_2 = GameLobby(
             TEST_USER_2, GameParameters(GameMode.RICHEST_PLAYER, 10, 25,
-            3600, 20, emptyList(), 2000), "Joinable 2", "lobby123abc"
+            3600, 20, getZones(), 2000), "Joinable 2", "lobby123abc"
         )
         val TEST_GAME_LOBBY_AVAILABLE_3 = GameLobby(
             TEST_USER_3, GameParameters(GameMode.RICHEST_PLAYER, 7, 77,
-            720, 20, emptyList(), 3000), "Joinable 3", "lobby1234abc"
+            720, 20, getZones(), 3000), "Joinable 3", "lobby1234abc"
         )
         val TEST_GAME_LOBBY_AVAILABLE_4 = GameLobby(
             TEST_USER_4, GameParameters(GameMode.RICHEST_PLAYER, 2, 4,
-            1080, 20, emptyList(), 4000), "Joinable 4", "lobbyabc1234"
+            1080, 20, getZones(), 4000), "Joinable 4", "lobbyabc1234"
         )
 
         val testPlayer1 = Player(TEST_USER_1, 100, listOf())
--- a/app/src/androidTest/java/com/github/polypoly/app/commons/PolyPolyTest.kt
+++ b/app/src/androidTest/java/com/github/polypoly/app/commons/PolyPolyTest.kt
@@ -23,6 +23,7 @@ import org.junit.After
 import org.junit.Before
 import org.junit.runner.RunWith
 import java.util.concurrent.CompletableFuture
+import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.locks.ReentrantLock
 import kotlin.math.absoluteValue
--- a/app/src/androidTest/java/com/github/polypoly/app/commons/PolyPolyTest.kt
+++ b/app/src/androidTest/java/com/github/polypoly/app/commons/PolyPolyTest.kt
@@ -18,10 +18,7 @@ import com.github.polypoly.app.utils.global.GlobalInstances.Companion.isSignedIn
 import com.github.polypoly.app.utils.global.GlobalInstances.Companion.remoteDB
 import com.github.polypoly.app.utils.global.GlobalInstances.Companion.remoteDBInitialized
 import com.google.firebase.auth.FirebaseAuth
-import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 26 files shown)


app/src/androidTest/java/com/github/polypoly/app/commons/PolyPolyTest.kt
```diff
@@ -69,27 +73,27 @@ abstract class PolyPolyTest(
 
         val TEST_GAME_LOBBY_FULL = GameLobby(
             TEST_USER_0, GameParameters(GameMode.RICHEST_PLAYER, 2, 6,
-            60, 20, emptyList(), 100), "Full gameLobby", "lobby1234"
+            60, 20, getZones(), 100), "Full gameLobby", "lobby1234"
```
tamaraaPrs, line 76: You do not need to specify the `gameMap` parameter when it contains all the zones because it is by default initialized to `getZone()`

app/src/androidTest/java/com/github/polypoly/app/map/VisitMapActivityTest.kt
```diff
@@ -78,13 +77,15 @@ class VisitMapActivityTest {
 
     @Test
     fun clickOnCloseButtonCloseTheDialog() {
-        val marker = MapUI.mapView.overlays.first { it is Marker} as Marker
-        composeTestRule.activity.mapViewModel.selectedMarker = marker
-        composeTestRule.activity.interactingWithProperty.value = true
+        setLocationInMapViewModel(getRandomLocation()).get(TIMEOUT_DURATION, TimeUnit.SECONDS)
         runBlocking { delay(500) }
```
tamaraaPrs, line 81: for all these `runBlocking { delay(500) }` you can use the function `waitForUIToUpdate` that @AmLibra do in `GameActivityTest` and put this function in a file in the package `commons`

app/src/main/java/com/github/polypoly/app/models/game/GameViewModel.kt
```diff
@@ -43,10 +63,160 @@ class GameViewModel(
...
+                remoteDB.setValue(gameUpdated)
+            } else { // up to date version already on live db
+                gameData.value = gameFound
+                CompletableFuture.completedFuture(true)
+            }
+        }
+    }
+
+    private fun playerStateFSMTransition(expectedFrom: PlayerState, to: PlayerState) {
+        if (playerStateData.value != expectedFrom)
```
tamaraaPrs, line 119: The case where the current player state in playerStateData is different from expectedFrom is an error that should never have happened? Because if so, maybe throw an exception instead

app/src/main/java/com/github/polypoly/app/models/game/GameViewModel.kt
```diff
@@ -43,10 +63,160 @@ class GameViewModel(
...
+
+            result.complete(closestLocation)
+        }
+
+        return result
+    }
+
+    /**
+     * Rolls the dice and returns the location that corresponds to the sum of 2 dice rolls, 3 times
+     * ensuring that the player does not visit the same location twice.
```
tamaraaPrs, line 192: You just forgot to specify what the param `currentLocation` was here in the comment!

app/src/main/java/com/github/polypoly/app/models/game/GameViewModel.kt
```diff
@@ -43,10 +63,160 @@ class GameViewModel(
...
+
+        viewModelScope.launch {
+            val locationsNotToVisitName = mutableListOf<String>()
+            if (currentLocation != null)
+                locationsNotToVisitName.add(currentLocation.name)
+
+            val allLocations = gameData.value?.allLocations ?: listOf()
+
+            val locationsToVisit = mutableListOf<LocationProperty>()
+            for (i in 1..3) {
```
tamaraaPrs, line 205: Maybe we should put global constants specific to the whole game and that one of these constants would be `numberOfLocationProposePerTurn = 3`

app/src/main/java/com/github/polypoly/app/models/game/GameViewModel.kt
```diff
@@ -43,10 +63,160 @@ class GameViewModel(
...
+                val diceRollsSum = IntArray(2) { (1..6).random() }.sum() - 2
```
tamaraaPrs, line 206: Same, maybe we can define constants like that: `numberOfDiceSides = 6` and `numberOfDices = 2`

app/src/main/java/com/github/polypoly/app/ui/game/PropertyInteractUIComponent.kt
```diff
@@ -9,84 +9,99 @@ import androidx.compose.material.AlertDialog
...
+ * @param gameViewModel GameViewModel to use for game business logic
+ * @param mapViewModel GameViewModel to use for map business logic
  */
 @Composable
-private fun PropertyInteractButtons(showBuyDialog: MutableState<Boolean>) {
+private fun PropertyInteractButtons(gameViewModel: GameViewModel, mapViewModel: MapViewModel) {
     Row(
         modifier = Modifier
             .fillMaxWidth()
             .padding(8.dp),
```
tamaraaPrs, line 82: I know it's not your code but we can replace that by `Padding.medium`

tamaraaPrs (changes requested):
Excellent PR, the code is mostly super clean! I only noticed a few small details but overall it was really great.

tamaraaPrs (changes requested):
Excellent PR, the code is mostly super clean! I only noticed a few small details but overall it was really great.

app/src/androidTest/java/com/github/polypoly/app/commons/PolyPolyTest.kt
Jharaxus: Indeed I didn't notice that, however if I only exclude this parameter it gives
`kotlin
GameParameters(GameMode.RICHEST_PLAYER, 4, 6, 360, 20,  initialPlayerBalance =  300)
`
which I find quite ugly.
Moreover, `GameMode.RICHEST_PLAYER` is actually also the default value but it is kept here as it is more meaningful.
Finally, as it is an instance specific for tests I prefer a being a bit more verbose to explicit its intent.

app/src/androidTest/java/com/github/polypoly/app/map/VisitMapActivityTest.kt
Jharaxus: Yep it is definitely clearer to do that !

app/src/main/java/com/github/polypoly/app/models/game/GameViewModel.kt
Jharaxus: I initially didn't want to have such a hard end upon wrong transitions, but it should indeed invalidate the fsm a bit more than by just ignoring what happened

app/src/main/java/com/github/polypoly/app/models/game/GameViewModel.kt
Jharaxus: I actually simplified the code because it doesn't really make sense to simulate real dice here

app/src/main/java/com/github/polypoly/app/ui/game/PropertyInteractUIComponent.kt
```diff
@@ -9,84 +9,99 @@ import androidx.compose.material.AlertDialog
...
         horizontalArrangement = SpaceEvenly
     ) {
         Button(
-            onClick = { showBuyDialog.value = true },
+            onClick = { gameViewModel.startBetting() },
```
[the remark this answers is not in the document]
Jharaxus: You're right, I also changed the enum constant name while I was at it

app/src/main/java/com/github/polypoly/app/ui/game/PropertyInteractUIComponent.kt
```diff
@@ -9,84 +9,99 @@ import androidx.compose.material.AlertDialog
...
 
-private fun leaveBetDialog() {
-    interactingWithProperty.value = false
-    mapViewModel.currentPlayer?.playerState?.value = PlayerState.INTERACTING
+private fun leaveInteractionDialog(gameViewModel: GameViewModel, mapViewModel: MapViewModel) {
+    gameViewModel.cancelBetting()
+    mapViewModel.selectLocation(null)
 }
+
+private fun onBuy(valueBet: Float, gameViewModel: GameViewModel) {
```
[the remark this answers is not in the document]
Jharaxus: Yes that's the work of a future pr that will complete the game loop, but I stopped there as the pr was already long enough (and took me enough time 😅)

app/src/androidTest/java/com/github/polypoly/app/commons/PolyPolyTest.kt
tamaraaPrs: Since we use "magic numbers", specifying all the parameters is not a bad idea either

tamaraaPrs (approved):

app/src/androidTest/java/com/github/polypoly/app/models/game/GameViewModelTest.kt
```diff
@@ -0,0 +1,195 @@
...
+        val model = GameViewModel(testGame, testPlayer, coroutineScope = this)
+
+        val positionOut = GeoPoint(0.toDouble(), 0.toDouble())
+        val locationFound = model.computeClosestLocation(positionOut).get(TIMEOUT_DURATION, TimeUnit.SECONDS)
+
+        assertNull(locationFound)
+    }
+
+    @Test
+    fun nexTurnSyncsDB() {
```
SolalPirelli, line 161: typo: `next`

app/src/androidTest/java/com/github/polypoly/app/models/game/GameViewModelTest.kt
```diff
@@ -0,0 +1,195 @@
...
+        model.nextTurn()
+        waitForUIToUpdate()
+
+        val updatedRound = remoteDB.getValue<Game>(TEST_GAME_LOBBY_AVAILABLE_2.key).get(TIMEOUT_DURATION, TimeUnit.SECONDS).currentRound
+
+        assertEquals(oldRound + 1, updatedRound)
+    }
+
+    @Test
+    fun nexTurnDoesNotSyncsDBIfUpToDateGameIsPresent() {
```
SolalPirelli, line 177: image

app/src/main/java/com/github/polypoly/app/base/game/PlayerState.kt
```diff
@@ -1,11 +1,11 @@
-package com.github.polypoly.app.ui.game
+package com.github.polypoly.app.base.game
 
 enum class PlayerState {
     INIT, // initial state, before the game starts
     ROLLING_DICE,
     MOVING,
     INTERACTING,
-    BETTING,
+    BIDDING,
```
SolalPirelli, line 8: 👍

app/src/main/java/com/github/polypoly/app/models/commons/LoadingModel.kt
```diff
@@ -24,5 +27,25 @@ abstract class LoadingModel: ViewModel() {
...
+        }
+    }
+
+    /**
+     * Registers a callback that will be called next time a data is synchronized with the storage
+     * This function is primarily but not exclusively aimed for used in unit tests
+     * @return a promise that completes when the data is synced again with the storage
+     */
+    fun waitForSync(): CompletableFuture<Boolean> {
+        return if (isLoading.value == false) {
```
SolalPirelli, line 44: this is odd to read, you could remove `== false` and invert the if/else bodies...

app/src/main/java/com/github/polypoly/app/models/commons/LoadingModel.kt
```diff
@@ -24,5 +27,25 @@ abstract class LoadingModel: ViewModel() {
      */
     protected fun setLoading(loading: Boolean) {
         isLoading.postValue(loading)
+        if (isLoading.value == true && !loading) {
```
SolalPirelli, line 30: no need for `== true`?

app/src/main/java/com/github/polypoly/app/ui/map/MapViewModel.kt
```diff
@@ -2,14 +2,15 @@ package com.github.polypoly.app.ui.map
```
SolalPirelli, line 2: since this file no longer depends on UI stuff, it could be moved to the package where it should be instead of `ui`?

PR #164, opened 2023-05-18
Extract hardcoded strings into string resources

jadtl:
This refactor extracts all hardcoded strings in the game into string resources to facilitate game localization down the road.
The main issue was extracting strings from enum classes and in some class functions, since getting the string from a resource ID requires a context object, I used two solutions respectively:

- Store the ID of the resource instead of a string in the enum classes.
- Pass a context object to the function that needs to get a resource string.


Kaz-ookid (approved):
nice work! This will greatly improve our code to be resilient and clear. Let's tell everyone to create a string constant anytime they hardcode a string for printing from now on!

jadtl:
The refactor was too tough to merge without issues last sprint so I'll close this PR for now.

PR #166, opened 2023-05-19
Implement passive capital gain in landlord mode

jadtl:
This implements a tax service that the game activity launches in the background to keep track of the player's location and tax them if they come near a location owned by another player.

The app requires background location permission that it requests to the user, and starts the service if granted, otherwise it acknowledges that the user declined to grant permission.

There are constants for the polling rate of the tax service and the distance to which it is considered that the player interacts with a building.

### DIFF
--- a/app/src/main/java/com/github/polypoly/app/base/game/service/LocationService.kt
+++ b/app/src/main/java/com/github/polypoly/app/base/game/service/LocationService.kt
@@ -1,0 +1,94 @@
+package com.github.polypoly.app.base.game.service
+
+import android.Manifest
+import android.app.Service
+import android.content.Intent
+import android.content.pm.PackageManager
+import android.location.Location
+import android.os.Build
+import android.os.IBinder
+import android.os.Looper
+import androidx.core.app.ActivityCompat
+import com.github.polypoly.app.utils.Constants
+import com.google.android.gms.location.FusedLocationProviderClient
+import com.google.android.gms.location.LocationCallback
+import com.google.android.gms.location.LocationRequest
+import com.google.android.gms.location.LocationResult
+import com.google.android.gms.location.LocationServices
+
+/**
+ * A service that runs in the background, receives location updates and processes them
+ */
+abstract class LocationService: Service() {
+    private lateinit var fusedLocationProviderClient: FusedLocationProviderClient
+
+    private val locationCallback = object : LocationCallback() {
+        override fun onLocationResult(locationResult: LocationResult) {
+            locationResult.lastLocation?.let { location ->
+                // Process the received location update
+                processLocationUpdate(location)
+            }
+        }
+    }
+
+    override fun onCreate() {
+        super.onCreate()
+        fusedLocationProviderClient = LocationServices.getFusedLocationProviderClient(this)
+    }
+
+    override fun onStartCommand(intent: Intent?, flags: Int, startId: Int): Int {
+        startLocationUpdates()
+        return START_STICKY
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        fusedLocationProviderClient.removeLocationUpdates(locationCallback)
+    }
+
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/github/polypoly/app/base/game/service/TaxService.kt
```diff
@@ -0,0 +1,37 @@
...
+ * if they are in a [LocationProperty] that is owned by another [Player].
+ */
+class TaxService: LocationService() {
+    override fun processLocationUpdate(location: Location) {
+        val playerLocation = location // For clarity
+
+        // Iterate through all the locations and check if the player is close enough to interact
+        for (inGameLocation in GameRepository.game?.inGameLocations!!) {
+            // Create a location object from the in-game location
+            val propertyLocation = Location("").apply {
```
tamaraaPrs, line 18: I think that instead of creating a location that is only used to calculate the distance, you could add in `LocationProperty` a function `distanceTo(location: Location)`  that do the same things, because I think we need to compute this distance many times in the code!

tamaraaPrs (changes requested):
The code is very clean, thank you! Only tiny little changes :)

tamaraaPrs (approved):

app/src/main/java/com/github/polypoly/app/base/game/service/LocationService.kt
```diff
@@ -0,0 +1,94 @@
...
+        // Check that location permissions are granted
+        if (ActivityCompat.checkSelfPermission(
+                this,
+                Manifest.permission.ACCESS_FINE_LOCATION
+            ) != PackageManager.PERMISSION_GRANTED && ActivityCompat.checkSelfPermission(
+                this,
+                Manifest.permission.ACCESS_COARSE_LOCATION
+            ) != PackageManager.PERMISSION_GRANTED
+        ) {
+            return
```
SolalPirelli, line 65: Should this component communicate errors somehow rather than dropping location updates on the floor and hoping nobody notices? :-)

app/src/main/java/com/github/polypoly/app/base/game/service/LocationService.kt
```diff
@@ -0,0 +1,94 @@
...
+            interval = Constants.LANDLORD_TAX_POLL_RATE * 1000 // Update interval in milliseconds
+            fastestInterval = Constants.LANDLORD_TAX_POLL_RATE * 1000 // Fastest update interval in milliseconds
+            priority = LocationRequest.PRIORITY_HIGH_ACCURACY
+        }
+
+        // Check that location permissions are granted
+        if (ActivityCompat.checkSelfPermission(
+                this,
+                Manifest.permission.ACCESS_FINE_LOCATION
+            ) != PackageManager.PERMISSION_GRANTED && ActivityCompat.checkSelfPermission(
```
SolalPirelli, line 60: weird formatting

app/src/main/java/com/github/polypoly/app/base/game/service/TaxService.kt
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * A service that runs in the background, receives location updates and taxes the [Player]
+ * if they are in a [LocationProperty] that is owned by another [Player].
+ */
+class TaxService: LocationService() {
+    override fun processLocationUpdate(location: Location) {
+        val playerLocation = location
+
+        // Iterate through all the locations and check if the player is close enough to interact
+        for (inGameLocation in GameRepository.game?.inGameLocations!!) {
```
SolalPirelli, line 16: these `?` and `!!` seem like symptoms of a (preexisting) data modelling issue

SolalPirelli:
This is a fine implementation to start with; if you evolve this app further after the course, FYI, this concept is called "geofencing" and modern OSes have some support for it https://developer.android.com/training/location/geofencing

PR #175, opened 2023-05-23
Add more context UI for current and Going to Location.

AmLibra:
Also fixed a small bug where the location would not update if you leave a location and only if you actually reached another location.


Kaz-ookid (approved):
looks nice and clear, nice work :)!

PR #172, opened 2023-05-22
added rules content

Kaz-ookid:
added rules content :D

PR #179, opened 2023-05-23
Hide round duration and rounds number selectors when Last standing

Kaz-ookid:
this PR makes so round duration and rounds number selectors are not visible when selecting last standing game mode because last standing game mode is not round based


vrooomer (approved):
Perfect!! I was talking about that in your PR #172, now it is way more intuitive, nice job!

~ v:)

PR #180, opened 2023-05-23
Fix HUD layout

jadtl:
This fixes the HUD layout where its elements were not correctly constrained, and the location indicator being displayed on top of the drop down menu when it drops down.

### DIFF
--- a/app/src/main/java/com/github/polypoly/app/ui/game/HeadsUpDisplay.kt
+++ b/app/src/main/java/com/github/polypoly/app/ui/game/HeadsUpDisplay.kt
@@ -6,6 +6,7 @@ import androidx.compose.foundation.border
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/polypoly/app/ui/game/HeadsUpDisplay.kt
```diff
@@ -66,22 +68,19 @@ fun Hud(
...
-                "Going to: "
-            )
-        HudGameMenu()
-    }
+    HudLocation(location, testTag = "interactable_location_text")
+    if (playerState == PlayerState.MOVING)
+        HudLocation(
+            playerPosition,
+            DpOffset(0.dp, 80.dp),
+            "going_to_location_text",
```
tamaraaPrs, line 78: here I would specify the name of the parameters for readability, to know which string do what

tamaraaPrs (changes requested):
Good! Just a very small change for the readability :)

tamaraaPrs (approved):

PR #181, opened 2023-05-23
Add a way to update Markers during the game

AmLibra:
This is going to be used to update the UI on turn switches and to show owned locations and such.


app/src/main/java/com/github/polypoly/app/base/game/location/LocationProperty.kt
```diff
@@ -17,7 +17,7 @@ data class Zone(
     An enum class representing the different levels of a [LocationProperty].
  */
 enum class PropertyLevel {
-    LEVEL_0, LEVEL_1, LEVEL_2, LEVEL_3, LEVEL_MAX
+    LEVEL_0, LEVEL_1, LEVEL_2, LEVEL_MAX
```
tamaraaPrs, line 20: We will have to think about replacing them with our real level names. Why did we delete a level by the way?

app/src/main/java/com/github/polypoly/app/ui/map/MapInitializationHelpers.kt
```diff
@@ -77,33 +81,78 @@ fun initMapView(context: Context): MapView {
  * @param gameViewModel GameViewModel to use for game business logic. Null if no game is going on.
  * @return the marker that was added
  */
-fun addMarkerTo(mapView: MapView, location: LocationProperty, zoneColor: Int,
-                mapViewModel: MapViewModel, gameViewModel: GameViewModel?
+fun addMarkerTo(
```
tamaraaPrs, line 84: I find it cleaner to put one parameter per line for readability

app/src/main/java/com/github/polypoly/app/ui/map/MapUI.kt
```diff
@@ -1,40 +1,69 @@
...
+                if (gameViewModel != null)
+                    mapViewModel.updateMapViewState(gameViewModel.getGameData().value!!.inGameLocations)
+                val mapView = createMapView(context, mapViewModel, gameViewModel, mapViewState)
                 mapView
-            }, modifier = Modifier.testTag("map"))
+            }, modifier = Modifier.testTag("map")
+        )
+    }
+
+    private fun createMapView(
```
tamaraaPrs, line 34: same, it is more lisible to put one parameter per line

tamaraaPrs (changes requested):
Everything is good except for the very small readability details!

app/src/main/java/com/github/polypoly/app/base/game/location/LocationProperty.kt
AmLibra: Reason 1, convenience and making the stars fit on the bitmap, I can make 3 stars fit ish (technically only 2 fit but conveniently 3 shows a full star and 2 star halves on either end)
Reason 2, we will never reach 4 levels anyway.

app/src/main/java/com/github/polypoly/app/ui/map/MapInitializationHelpers.kt
AmLibra: done

app/src/main/java/com/github/polypoly/app/ui/map/MapUI.kt
AmLibra: done

tamaraaPrs (approved):

PR #184, opened 2023-05-24
Refactoring

tamaraaPrs:
In this PR:

- add some comments or improve some existing ones
- rename all _bet_ in _bid_
- rename some `testTag` to respect our testTag namming covention (snake_case)
- delete the `currentGame` instance in the class `Game` and change all the `Game.currentGame` with `GameRepository.game`
- rename the package `models` to `viewmodels`
- rename the package `network` to `database`
- rename the test package `map` to `game_and_map`
- delete some unuse and unuseful code
- other small changes


app/src/main/java/com/github/polypoly/app/ui/menu/SignInActivity.kt
```diff
@@ -214,11 +211,11 @@ class SignInActivity : ComponentActivity() {
             trophiesDisplay = mutableListOf(0, 4)
         )
 
-        val TEST_USER_1 = User("12","Carter", "Not me!", NO_SKIN, ZERO_STATS, listOf(), mutableListOf())
-        val TEST_USER_2 = User("123","Harry", "Ha!", NO_SKIN, ZERO_STATS, listOf(), mutableListOf())
-        val TEST_USER_3 = User("1234","James", "Hey!", NO_SKIN, ZERO_STATS, listOf(), mutableListOf())
-        val TEST_USER_4 = User("12345","Henri", "Ohh!", NO_SKIN, ZERO_STATS, listOf(), mutableListOf())
-        val TEST_USER_5 = User("123456", "test_user_5", "", NO_SKIN, ZERO_STATS, listOf(), mutableListOf())
+        val TEST_USER_1 = User("12","Carter", "Not me!", Skin.default(), Stats())
```
Jharaxus, line 214: Why is there a Skin.default() but not a Stats.default() as well? Actually all our data classes have a default constructor and I don't think they need a `.default` function as they already have default values in their constructors for that

Jharaxus (changes requested):
Nice refactoring, some cleanup was indeed needed!
Two small additional notes to the little comment I put:
- the pr's name has a typo
- I believe that when doing a refactoring pr such as that one where you move a lot of files and change a lot of names, it is better to not include actual code logic refactoring as it may introduce bugs hard to find as this is not the initial purpose of this pr, and this can easily make resolving the resulting conflicts with other branches a lot harder

app/src/main/java/com/github/polypoly/app/ui/menu/SignInActivity.kt
tamaraaPrs: `Skin.default()` because there are a "default" skin that all users have at the creation of their acount but it is not necessary `Skin(0,0,0)`, it can be `Skin(1,3,2)` for example, it is just the skin when the player haven't choose their Skin yet. There are no "default" Stat because by definition the stat begin at 0, it can't be ambigus. I think `Skin.defaukt()` is clearer than `Skin()` because it indicate that we _choose_ a default Skin, that it is deliberate and not just to put some default values for the DB.

Jharaxus: Ok that makes sense! Then maybe juste put it as a variable instead of a function and it should be good!

Jharaxus (approved):
LGTM !

PR #187, opened 2023-05-25
Transaction enum and bill attribute in player

Kaz-ookid:
created `Transaction ` enum to ease money earnings and losing computatin, and summing up transactions happened during a round.

Transactions are divided in two categories: the ones that are computed in live, and the ones that wait for the end of the round to be computed. In both case, the `bill `attribute represents a way to keep track of what was done, and was must be done between two rounds.

I decided not to adapt the rest of the app with it to keep this PR short and quickly merged.

To adapt: make so that any money earnings and losings pass by filling up the `bill `attribute. The `bill `attribute will have to be read and cleared at the end of rounds to compute the players total gains during the round.


app/src/main/java/com/github/polypoly/app/base/game/Transaction.kt
```diff
@@ -0,0 +1,53 @@
...
+ */
+enum class Transaction(private val title : String, private val duringRound: Boolean, private val gameModes : List<GameMode>) {
+    TAX_TO_PAY ("Taxes to pay:", false, listOf(GameMode.RICHEST_PLAYER, GameMode.LAST_STANDING)),
+    TAX_TO_RECEIVE("You taxed players: ", false, listOf(GameMode.RICHEST_PLAYER, GameMode.LAST_STANDING)),
+    MILESTONES_INCOME_TO_RECEIVE("Milestones reward: ", false, GameMode.values().toList()),
+    PASSIVE_LANDLORD_TAX_PAID("Passive taxes to pay: ", true, listOf(GameMode.LANDLORD)),
+    PASSIVE_LANDLORD_TAX_RECEIVED("You taxed passively players: ", true, listOf(GameMode.LANDLORD)),
+    TRADE_MONEY_PAID("Spent in trades: ", true, GameMode.values().toList()),
+    TRADE_MONEY_RECEIVED("Earned in trades: ", true, GameMode.values().toList()),
+    BUILDING_BOUGHT_PAID("Spent in buildings: ", true, GameMode.values().toList());
```
Jharaxus, line 16: The messages don't each reveal the same intent: for example, `Taxes to pay:` tells that the taxes are not payed yet, while `You taxed players:` tells that the taxes are already payed, is it intended?

Jharaxus: Moreover, this kind of strings are not really titles but descriptions that should be moved to the UI instead.

app/src/main/java/com/github/polypoly/app/base/game/Player.kt
```diff
@@ -18,6 +18,7 @@ data class Player (
     val user: User = User(),
     private var balance: Int = 0,
     private var roundLost: Int? = null,
+    private val bill : Set<Transaction> = Transaction.values().toSet(),
```
Jharaxus, line 21: I don't think this is the best place to put this: as it is a turn-specific value it would better fit in Game as this notion of turn is not really in the conceptual space of the player: it doesn't know in which turn it is, not wether is should clear its bill or not, whereas `Game` does know that and is responsible for game turns' consistency

Jharaxus (changes requested):
Nice idea, this looks promising to have more modularity in our code!
However, I am not entirely convinced by how it looks right now: it looks well designed to have a nice summary of the previous turn's money flows at the beginning of the new one, as it condenses them into aggregated transactions bundled by type, but it is not very versatile for the rest of the game's timeline.
What I mean here is that it cannot really be used outside of this goal, and therefore still needs additional logic for other cases and sometimes leads to duplicating information in the code: for example, when having to pay a tax for a building we went through, it would be nice to also know which building caused that tax, instead of only the amount, which is not possible here without having a second list of buildings that the player went through. In the same idea, milestones could also benefit from having the number of meters walked together with the money earned for that, but it is also not possible here without quite more extra logic.
Also, with the bidding feature, the entire bidding transaction already needs to be saved to know which building was bid for, which includes the money flow. This therefore creates a duplicate with the `BUILDING_BOUGHT_PAID` transaction. Then, when should we remove the player’s money for a bid? When processing `BUILDING_BOUGHT_PAID` as with all other transactions? But then it is separated from the logic of processing bids (even though we decided to remove the bid money anyway) which is not very nice to do… But if we remove the money during the bidding processing, then the transaction processing becomes inconsistent as some are processed during turn, some at the end and some at the end but not with the others…
I also feel like the `duringRound` parameter reveals that this class doesn’t have a clearly defined intent: if a transaction is processed during the round, then what’s left is more a partial transaction summary log then an actual transaction… And what’s processed in that case is not the transaction itself, but several transactions whose values are added in this log (otherwise you would process the first transactions several times)…
Finally, the `TRADE_MONEY_PAID` and `TRADE_MONEY_RECEIVED` are nice but trades don't involve money as of now, so I believe this also shows that this money transaction concept is a bit too restrictive right now…
However, I still believe this is an interesting concept that we can bit adapted bit to be even nicer:
What you could do is simply generalize the concept money transaction to « this is a transaction initiated by a player that involves something ». This could for example take the form of an abstract class storing the title / player involved / wether is is already proceeded or not, and then each transaction in the game, a trade, a bid, a tax, or a money gift for steps walked, would be childs of this Transaction type (so trade request could for example have an additional `TradeTransaction` created for each player to represent a successful trade, and `LocationBid` could become `BidTransaction` as it already fits this general transaction concept). Then, each transaction could either be processed when created, or at the end of the turn where all unprocessed transactions are processed. Then you could get a summary of what happened during a turn by aggregating all transactions by kind (= type), while keeping each of their individual information if needed for the UI to display. Moreover, it would now permit having all of each transaction’s logic in the same place, which therefore solves the issue of duplicating the money logic and not knowing where to process it. Finally, this would give some coherence between the existing trades and create a smoother interface for the gameViewModel to expose as everything that affects the game would be a transaction —> the logic would be simpler and easier to maintain
Finally the end, this is a huge description 😅
I am of course open to discussion if you don’t agree with any of my points or with the solution I proposed :)

app/src/main/java/com/github/polypoly/app/base/game/Transaction.kt
Kaz-ookid: you are right. I will make these more clear. Also I will change them into real titles, and have the formatting part in the toString function, as it is a much more logical way to go.

app/src/main/java/com/github/polypoly/app/base/game/Transaction.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * This function is used to know if the transaction is compatible with the game mode
+     */
+    fun isGameModeCompatible(gameMode: GameMode) : Boolean {
+        return gameModes.contains(gameMode)
+    }
+
+    override fun toString(): String {
+        return "$title$amount $"
```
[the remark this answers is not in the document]
Kaz-ookid: It was in the title part, but it will be changed as said in the other comment  :)

app/src/androidTest/java/com/github/polypoly/app/base/game/GameTest.kt
```diff
@@ -146,4 +147,54 @@ class GameTest {
                 }
         }
     }
+
+    @Test
+    fun whenNextTurnComputesTransactions() {
+
+        gameLobby.start()
+        val game = Game.gameInProgress!!
```
Jharaxus, line 155: `gameLobby.start()` returns the game created that you can use instead of `Game.gameInProgress!!`

app/src/androidTest/java/com/github/polypoly/app/base/game/GameTest.kt
```diff
@@ -146,4 +147,54 @@ class GameTest {
...
+        game.nextTurn()
+
+        assert(game.transactions.isEmpty())
+        for (transactions in game.pastTransactions){
+            for (transaction in transactions) {
+                assertTrue(transaction.isExecuted())
+            }
+        }
+
+        assert(player1MoneyBefore + 100 != player1.getBalance())
```
Jharaxus, line 197: Shouldn't it be `==` instead of `!=`

app/src/main/java/com/github/polypoly/app/base/game/Game.kt
```diff
@@ -32,6 +30,8 @@ class Game private constructor(
     val dateBegin: Long = System.currentTimeMillis(),
     val inGameLocations: List<InGameLocation> = rules.gameMap
         .flatMap { zone -> zone.locationProperties.map { InGameLocation(it) } },
+    val transactions : ArrayList<GameTransaction> = arrayListOf(),
+    val pastTransactions: ArrayList<List<GameTransaction>> = arrayListOf(),
```
Jharaxus, line 34: Maybe you could have a `Map<Int, List<GameTransaction>>` instead to better express that each list is for one turn (and make it more easily usable if needed)

Kaz-ookid:
Thank you for this detailed and helpful review. You are right when saying that this this whole bill implementation is useless out of its purpose (summary of money ins and outs), because it was designed only for this and with no higher expectation. But I have to agree with you that it yields a very interesting idea of stretching the concept larger and think bigger, better. As you said, we can aim to keep a trace of all transactions as well as their whole context, rather than just the money value flow. 


`This therefore creates a duplicate with the BUILDING_BOUGHT_PAID transaction. Then, when should we remove the player’s money for a bid? When processing BUILDING_BOUGHT_PAID as with all other transactions? But then it is separated from the logic of processing bids (even though we decided to remove the bid money anyway) which is not very nice to do… But if we remove the money during the bidding processing, then the transaction processing becomes inconsistent as some are processed during turn, some at the end and some at the end but not with the others…`

For this part above, I must say that I forgot our own rules and thought that buildings were supposed to be bought as it is, not using bids. This is my mistake, not remembering that there are no such things as buying buildings without going through bids :)


`Finally, the TRADE_MONEY_PAID and TRADE_MONEY_RECEIVED are nice but trades don't involve money as of now, so I believe this also shows that this money transaction concept is a bit too restrictive right now…`

For this part, once again I did not see that trades could not involve money, that is all :) !


In conclusion, I decided to go backwards and deleted Transactions to create GameTransactions abstract class, that holds common transactions mechanismes. I decided not to implement the subclasses (such that BiddingTransaction, TaxTransaction or any other we can think about...) because I find it more relevant to get this global structure merged quickly, so that we can implement these over time and separately (for example, as we talked you will be implementing BiddingTransaction and I will implement MilestoneRewardTransaction).

Jharaxus (changes requested):
Great new class, that offers nice opportunities for a clean code! The abstraction is well thought too!

app/src/androidTest/java/com/github/polypoly/app/base/game/GameTest.kt
Kaz-ookid: you are right thanks :) !

app/src/androidTest/java/com/github/polypoly/app/base/game/GameTest.kt
Kaz-ookid: no, actually player1 balance is modified in the execution of the first mock transaction, that is supposed to be executed aready! So this player should in fact not have gained money after the transactions computing in game :)

app/src/main/java/com/github/polypoly/app/base/game/Game.kt
Kaz-ookid: thats right :)!

Jharaxus (approved):
LGTM !

PR #188, opened 2023-05-25
Final board UI and logic

### LINKED ISSUE #189: Add final winning board at end of game

vrooomer:
Small PR to implement a dialog that shows up whenever a game is finished.
The endscreen contains:
- A message "End of the game!"
- A ranking of the players
- A button to quit the game and return to the menu (this one will set every static game information to `null`)

The button will also update the player's stats (i.e. number of played games and number of won games)

_Note: still draft because the tests aren't finished_

_Edit: now the PR is open because the tests are finished, some are sadly **commented**, check the discussion for more information about that_


app/src/main/java/com/github/polypoly/app/ui/game/GameEndUIComponents.kt
```diff
@@ -0,0 +1,142 @@
...
+            Column(
+                modifier = Modifier
+                    .fillMaxWidth(0.9f)
+                    .fillMaxHeight(0.5f),
+                horizontalAlignment = Alignment.CenterHorizontally,
+                verticalArrangement = Arrangement.Center
+            ) {
+                GameBoard(getPlayersInRankingOrder(lastGame))
+                Spacer(modifier = Modifier.height(Padding.large))
+                UIElements.BigButton(onClick = quitAction, text = "Return to menu")
```
jadtl, line 55: You can extract this string as a string ressource :)

app/src/main/java/com/github/polypoly/app/ui/game/GameEndUIComponents.kt
```diff
@@ -0,0 +1,142 @@
...
+    ) {
+        Column(
+            modifier = Modifier
+                .background(MaterialTheme.colors.primary)
+                .padding(Padding.large),
+            horizontalAlignment = Alignment.CenterHorizontally,
+            verticalArrangement = Arrangement.spacedBy(Padding.large)
+        ) {
+            Text(
+                text = "End of the game!",
```
jadtl, line 77: This one can also be extracted as a string ressource.

app/src/main/java/com/github/polypoly/app/ui/game/GameEndUIComponents.kt
```diff
@@ -0,0 +1,142 @@
...
+    }
+}
+
+@Composable
+private fun LeaderBoardRow(player: Player, rank: Int) {
+    var textColor = MaterialTheme.colors.onPrimary
+    var color = MaterialTheme.colors.secondary
+
+    if(player.user.currentUser) {
+        color = MaterialTheme.colors.onSecondary
```
jadtl, line 104: `color` being a background color, why does it use the color `onSecondary` instead of `secondary`?

app/src/main/java/com/github/polypoly/app/ui/game/GameEndUIComponents.kt
```diff
@@ -0,0 +1,142 @@
...
+            for (i in orderedPlayers.indices) {
+                LeaderBoardRow(player = orderedPlayers[i], rank = i+1)
+            }
+        }
+    }
+}
+
+@Composable
+private fun LeaderBoardRow(player: Player, rank: Int) {
+    var textColor = MaterialTheme.colors.onPrimary
```
jadtl, line 100: Why is `textColor` `onPrimary` when `color` is a secondary color?

jadtl (changes requested):
Nicely done! Thanks for bringing this feature to wrap up the game UI :)

I have a few comments, feel free to go through them when you have time, thanks!

And good luck with tests!

app/src/main/java/com/github/polypoly/app/ui/game/GameEndUIComponents.kt
vrooomer: I tried but as this is not an activity, I don't have access to the usual `getString` function, let me know if you have any advice for that, I added the strings in the resources still :)

app/src/main/java/com/github/polypoly/app/ui/game/GameEndUIComponents.kt
vrooomer: same as above

app/src/main/java/com/github/polypoly/app/ui/game/GameEndUIComponents.kt
vrooomer: I had to find colors to differentiate odd and even lines and a different one for the row of the current user, as we don't have that much colors I chose that one :)

app/src/main/java/com/github/polypoly/app/ui/game/GameEndUIComponents.kt
vrooomer: You're right about that one, it was only an aesthetic choice, I can change it if you think that doesn't match

vrooomer:
Thanks for the review! Let me know if you still want me to change things!

I added the tests now, sadly three of them don't pass because of a time out in an external function, I tried to debug during two good hours but nothing...
Hopefully I was able to test this UI (they were UI tests) manually and it works perfectly! :)

Also, the timed out function `GameViewModel.nextTurn()` never timed out before in the app so it may be a bug coming directly from the way we do our tests.

~ v:)

app/src/main/java/com/github/polypoly/app/ui/game/GameEndUIComponents.kt
jadtl: It's only a nitpick since it's more of a design issue, so it's not urgent to fix it :)

app/src/main/java/com/github/polypoly/app/ui/game/GameEndUIComponents.kt
jadtl: I think that you can call the `stringResource` function to access string resources inside composables.

jadtl (approved):

PR #190, opened 2023-05-25
Bug/player storing in game locations

Jharaxus:
There is currently a bug where if a player owns a location then the Game object is not serializable anymore due to the circular references where `Player` had an attribute `ownedLocations` and these locations had an attribute `owner` referencing the player owning them.
It also removes the redundant list of `InGameLocation` owned that was stored in the `MapViewModel`, and adds instead a `mutableStateData` associated to it in `gameViewModel` as otherwise the UI couldn't detect changes in this list.


Kaz-ookid (approved):
Thank you for fixing this! This is a much needed fix. Also, nice initiative to move the tradeWith function to GameViewModel from Player, as this really should not take place their at all.

Nice work!  :)

PR #191, opened 2023-05-26
Update game UI

jadtl:
This updates the game UI so that every element complies with our design language, and so that the user can look at their detailed statistics as well as game statistics during the game.


app/src/main/java/com/github/polypoly/app/ui/game/HeadsUpDisplay.kt
```diff
@@ -365,6 +369,7 @@ fun HudButton(testTag: String, onClick: () -> Unit, icon: ImageVector, descripti
             .testTag(testTag)
             .padding(bottom = Padding.large),
         shape = CircleShape,
+        elevation = ButtonDefaults.elevation(0.dp),
```
Kaz-ookid, line 372: does an elevation of 0 does anything?

Kaz-ookid (approved):
Love the final look! This pr brings us great modularity for UI, nice job! However there is a large use of magic numbers, may be for the future we will have to browse the code and gather those under constants :) for now it will do

app/src/main/java/com/github/polypoly/app/ui/game/HeadsUpDisplay.kt
jadtl: Yes, it removes the drop down shadow from the buttons to give them, like the other UI elements, a flat design feel as @tamaraaPrs suggested we do 😊

jadtl:
> However there is a large use of magic numbers

These magic numbers were part of a previous PR, I've only modified them, but I agree that they should eventually be abstracted!

app/src/androidTest/java/com/github/polypoly/app/base/game/GameLobbyActivityTest.kt
```diff
@@ -163,27 +163,27 @@ class GameLobbyActivityTest: PolyPolyTest(true, false, true) {
...
+    //    composeTestRule.onNodeWithTag("game_lobby_leave_button", useUnmergedTree = true)
+    //        .performClick()
+
+    //    val gameLobbyAfterClick = composeTestRule.activity.gameLobbyWaitingModel.getGameLobby().value!!
+    //    assert(!gameLobbyAfterClick.usersRegistered.contains(currentUser))
+    //    assert(gameLobbyAfterClick.admin != currentUser)
+    //    assert(gameLobbyAfterClick.usersRegistered.count{it.id == gameLobbyAfterClick.admin.id} == 1)
+
+    //    resetGameLobby()
+    //}
```
SolalPirelli, line 186: hmmmm

PR #192, opened 2023-05-26
Update markers upon change and tax players when visiting a building

Jharaxus:
- Fix a bug that prevented markers from updating in the mapView
- Adds taxing players when they try to bid on a location that already belongs to someone else
- Adds random upgrade of one location per user at each turn


app/src/main/java/com/github/polypoly/app/base/game/location/InGameLocation.kt
```diff
@@ -47,4 +47,13 @@ data class InGameLocation (
...
     }
+
+    fun upgrade() {
+        level = when(level) {
+            PropertyLevel.LEVEL_0 -> PropertyLevel.LEVEL_1
+            PropertyLevel.LEVEL_1 -> PropertyLevel.LEVEL_2
+            PropertyLevel.LEVEL_2 -> PropertyLevel.LEVEL_MAX
+            PropertyLevel.LEVEL_MAX -> PropertyLevel.LEVEL_MAX
+        }
+    }
```
jadtl, line 58: You could use `values` and increment the array unless you're at the last index.

jadtl (approved):
Thanks for adding this feature, it seems to work well and your code is nice as well :)

Just a small nitpick I'll let you have a look at.

app/src/main/java/com/github/polypoly/app/base/game/location/InGameLocation.kt
Jharaxus: Yes that's right, however I didn't do that as the code with `values()`, `valuesOf()` and the check that we are not in the last value would be a bigger and less clear code in the end. Moreover `when` doesn't compile if not exhaustive so it shouldn't have weird behavior anyway...

vrooomer:
In this PR my PR will be merge too, check out #188 for more information :)

~ v:)

vrooomer (approved):
I won't review my code but for the other one, after the minor fixes that we discussed in person, it looks very clean now, great job :)

~ v:)

